PR #1, opened 2018-04-09
adds Circle CI config for initial builds

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2737.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,17 @@
+version: 2
+jobs:
+  build:
+    machine: true
+    steps:
+    - checkout
+
+    - run: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,17 @@
...
+    machine: true
+    steps:
+    - checkout
+
+    - run: |
+        wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
+        chmod +x ./architect
+        ./architect version
+    - run: ./architect build
+    - deploy:
```
fgimenez: It would be good to start having a separate job for deploying that is run after e2e on master, as in https://github.com/giantswarm/chart-operator/blob/master/.circleci/config.yml. Not a blocker, we can iterate on this later.

xh3b4sd: Ack. Will consider this in the next PRs.

PR #4, opened 2018-04-09
adds Integration tests

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2737.


.circleci/config.yml
```diff
@@ -1,3 +1,29 @@
...
+    - run: ./e2e-harness localkube
+
+    - run: ./e2e-harness setup --remote=false
+
+    - run: ./e2e-harness test --test-dir=integration/test/basic
+
+    - run:
+        name: Finish with cleanup, no matter if the test succeeded or not
+        command: ./e2e-harness teardown
+        when: always
```
xh3b4sd: So here I noticed we do not need the CNR server magic because the integration tests works against the fixures in the local FS. Is that about right?

integration/test/basic/project.yaml
```diff
@@ -0,0 +1,6 @@
+version: 1
+
+test:
+  env:
+    - "KEEP_RESOURCES=${KEEP_RESOURCES}"
+    - "CIRCLE_SHA1=${CIRCLE_SHA1}"
```
xh3b4sd: I think this can actually be removed. Can somebody confirm?

.circleci/config.yml
fgimenez: Yep, that's only needed for `apprclient`.

integration/test/basic/project.yaml
fgimenez: Exactly, same as above, only needed for CNR-related stuff

xh3b4sd: Sweet. Removed.

PR #9, opened 2018-04-12
provides primitives to install tiller in a k8s cluster

xh3b4sd:
This is towards https://github.com/giantswarm/giantswarm/issues/1902 and https://github.com/giantswarm/giantswarm/issues/2737. We have some overlap here but I took this now because I need it for the integration tests.


xh3b4sd:
Requesting review and awareness here. Need to test this though.

Gopkg.toml
```diff
@@ -26,6 +26,21 @@
...
+required = [
+  "github.com/Azure/go-autorest/autorest",
+	"github.com/docker/distribution/reference"
+]
+
+
+
+[[constraint]]
+  name = "github.com/Azure/go-autorest"
+  version = "9.1.0"
```
kopiczko: Is that needed?

Gopkg.toml
```diff
@@ -26,6 +26,21 @@
...
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/docker/distribution"
```
kopiczko: And that?

Gopkg.toml
xh3b4sd: Well yes, which is why I added it. At least I did not find a way to get the deps compiled without it. These things are always because of Kubernetes dpendencies. When you find a nicer more default way to compile this shit, let me know.

Gopkg.toml
xh3b4sd: See my comment above.

helmclient.go
```diff
@@ -116,68 +138,237 @@ func (c *Client) GetReleaseHistory(releaseName string) (*ReleaseHistory, error)
...
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
 	return nil
 }
 
+func (c *Client) InstallTiller() error {
+	var name = "tiller"
+	var namespace = "kube-system"
```
fgimenez: We have `tillerDefaultNamespace`. Would also be nice to have a constant for the name.

helmclient.go
```diff
@@ -116,68 +138,237 @@ func (c *Client) GetReleaseHistory(releaseName string) (*ReleaseHistory, error)
...
+			// fall through
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	// Install the tiller deployment in the guest cluster.
+	{
+		o := &installer.Options{
+			ImageSpec:      "gcr.io/kubernetes-helm/tiller:v2.8.2",
```
fgimenez: Constants here for image and tag would be nice too.

helmclient.go
xh3b4sd: Good shout. Done.

helmclient.go
xh3b4sd: Sorted.

PR #10, opened 2018-04-12
adds backoffs everywhere

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2737. While testing around I figured working against a CNR is not really reliable. Especially not when having port forwarding in between. This PR tries to harden the client by adding backoffs where necessary.

### DIFF
--- a/error.go
+++ b/error.go
@@ -7,9 +7,28 @@ import (
 )
 
 const (
-	releaseNotFoundErrorPrefix = "No such release:"
-	releaseNotFoundErrorSuffix = "not found"
+	cannotReuseReleaseErrorPrefix = "cannot re-use"
 )
+
+var cannotReuseReleaseError = microerror.New("cannot reuse release")
+
+// IsCannotReuseRelease asserts cannotReuseReleaseError.
+func IsCannotReuseRelease(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	c := microerror.Cause(err)
+
+	if strings.Contains(c.Error(), cannotReuseReleaseErrorPrefix) {
+		return true
+	}
+	if c == cannotReuseReleaseError {
+		return true
+	}
+
+	return false
+}
 
 var executionFailedError = microerror.New("execution failed")
 
@@ -27,10 +46,15 @@ func IsInvalidConfig(err error) bool {
 
 var podNotFoundError = microerror.New("pod not found")
 
-// IsNotFound asserts podNotFoundError.
-func IsNotFound(err error) bool {
+// IsPodNotFound asserts podNotFoundError.
+func IsPodNotFound(err error) bool {
 	return microerror.Cause(err) == podNotFoundError
 }
+
+const (
+	releaseNotFoundErrorPrefix = "No such release:"
+	releaseNotFoundErrorSuffix = "not found"
+)
 
 var releaseNotFoundError = microerror.New("release not found")
 
--- a/helmclient.go
+++ b/helmclient.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/helm/cmd/helm/installer"
 	"k8s.io/helm/pkg/chartutil"
 	helmclient "k8s.io/helm/pkg/helm"
+	hapiservices "k8s.io/helm/pkg/proto/hapi/services"
 )
 
 // Config represents the configuration used to create a helm client.
@@ -67,13 +68,28 @@ func New(config Config) (*Client, error) {
 
 // DeleteRelease uninstalls a chart given its release name.
 func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteOption) error {
-	t, err := c.newTunnel()
-	if err != nil {
-		return microerror.Mask(err)
-	}
-	defer c.closeTunnel(t)
+	o := func() error {
+		t, err := c.newTunnel()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		defer c.closeTunnel(t)
 
-	_, err = c.newHelmClientFromTunnel(t).DeleteRelease(releaseName, options...)
+		_, err = c.newHelmClientFromTunnel(t).DeleteRelease(releaseName, options...)
+		if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+
+		return nil
+	}
+	b := newExponentialBackoff(2*time.Minute, backoff.DefaultMaxInterval)
+	n := func(err error, delay time.Duration) {
+		c.logger.Log("level", "debug", "message", "failed deleting release")
+	}
+
+	err := backoff.RetryNotify(o, b, n)
 	if err != nil {
 		return microerror.Mask(err)
 	}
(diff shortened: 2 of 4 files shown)


error.go
```diff
@@ -7,10 +7,29 @@ import (
 )
 
 const (
-	releaseNotFoundErrorPrefix = "No such release:"
-	releaseNotFoundErrorSuffix = "not found"
+	cannotReuseReleaseErrorPrefix = "cannot re-use"
```
xh3b4sd: Doing the error string matching dance.

error.go
```diff
@@ -27,11 +46,16 @@ func IsInvalidConfig(err error) bool {
 
 var podNotFoundError = microerror.New("pod not found")
 
-// IsNotFound asserts podNotFoundError.
-func IsNotFound(err error) bool {
+// IsPodNotFound asserts podNotFoundError.
+func IsPodNotFound(err error) bool {
```
xh3b4sd: Recently I changed the error but forgot about the matcher. Shame on me.

error.go
```diff
@@ -27,11 +46,16 @@ func IsInvalidConfig(err error) bool {
...
 	return microerror.Cause(err) == podNotFoundError
 }
 
+const (
+	releaseNotFoundErrorPrefix = "No such release:"
+	releaseNotFoundErrorSuffix = "not found"
+)
+
```
xh3b4sd: I moved this to have the things related to an error together. I find that easier to grasp what the things are about.

helmclient.go
```diff
@@ -67,13 +68,28 @@ func New(config Config) (*Client, error) {
 
 // DeleteRelease uninstalls a chart given its release name.
 func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteOption) error {
-	t, err := c.newTunnel()
-	if err != nil {
-		return microerror.Mask(err)
+	o := func() error {
```
xh3b4sd: I wrapped everything in backoffs. Please check if the calls are right. The great unit testing we did here so far helped a lot sorting this out, especially with the error handling. Thank you for the good work so far.

helmclient.go
```diff
@@ -67,13 +68,28 @@ func New(config Config) (*Client, error) {
...
+		t, err := c.newTunnel()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		defer c.closeTunnel(t)
+
+		_, err = c.newHelmClientFromTunnel(t).DeleteRelease(releaseName, options...)
+		if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
```
xh3b4sd: So note that we have to use a permanent error here to tell the retry to stop since we actually only want to catch unexpected stuff like broken network connections. Here we cannot use error masking because of the backoff internals. When the backoff stops it unwraps the given error and we can use masking again.

helmclient_test.go
```diff
@@ -262,7 +262,7 @@ func Test_GetReleaseHistory(t *testing.T) {
 	}
 }
 
-func Test_InstallRelease(t *testing.T) {
+func Test_Client_InstallFromTarball(t *testing.T) {
```
xh3b4sd: I figured the name das not match the method we are testing here.

helmclient.go
```diff
@@ -270,7 +338,7 @@ func (c *Client) InstallTiller() error {
 
 			return nil
 		}
-		b := newExponentialBackoff(2 * time.Minute)
+		b := newExponentialBackoff(2*time.Minute, 5*time.Second)
```
xh3b4sd: We are using the default `maxInterval` everywhere but here because the tiller installation backoff is about checking the service is up while the other backoffs are about retrying on network errors. This is why this got configurable.

error.go
```diff
@@ -7,10 +7,29 @@ import (
 )
 
 const (
-	releaseNotFoundErrorPrefix = "No such release:"
-	releaseNotFoundErrorSuffix = "not found"
+	cannotReuseReleaseErrorPrefix = "cannot re-use"
 )
 
+var cannotReuseReleaseError = microerror.New("cannot reuse release")
```
tuommaki: `"cannot reuse...` vs. `"cannot re-use"` - should these match?

helmclient.go
```diff
@@ -85,17 +101,35 @@ func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteO
...
+		o := func() error {
+			t, err := c.newTunnel()
+			if err != nil {
+				return microerror.Mask(err)
+			}
+			defer c.closeTunnel(t)
+
+			resp, err = c.newHelmClientFromTunnel(t).ReleaseContent(releaseName)
+			if IsReleaseNotFound(err) {
+				return backoff.Permanent(releaseNotFoundError)
```
tuommaki: I think that additional context information of `releaseName` should be added to wrapped error here. It was present in earlier version. So maybe this should be:
```go
return backoff.Permanent(microerror.Maskf(releaseNotFoundError, releaseName))
```

helmclient.go
```diff
@@ -122,18 +156,35 @@ func (c *Client) GetReleaseContent(releaseName string) (*ReleaseContent, error)
...
+			defer c.closeTunnel(t)
 
-	resp, err := c.newHelmClientFromTunnel(t).ReleaseHistory(releaseName, helmclient.WithMaxHistory(1))
-	if IsReleaseNotFound(err) {
-		return nil, microerror.Maskf(releaseNotFoundError, releaseName)
-	} else if err != nil {
-		return nil, microerror.Mask(err)
+			resp, err = c.newHelmClientFromTunnel(t).ReleaseHistory(releaseName, helmclient.WithMaxHistory(1))
+			if IsReleaseNotFound(err) {
+				return backoff.Permanent(releaseNotFoundError)
```
tuommaki: Same as above: maybe mask `releaseNotFoundError` with `releaseName`?

helmclient.go
```diff
@@ -158,13 +209,30 @@ func (c *Client) GetReleaseHistory(releaseName string) (*ReleaseHistory, error)
...
+	o := func() error {
+		t, err := c.newTunnel()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		defer c.closeTunnel(t)
+
+		_, err = c.newHelmClientFromTunnel(t).InstallRelease(path, ns, options...)
+		if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
```
tuommaki: Would additional context information of release name help in this error and also one below?

helmclient.go
```diff
@@ -289,13 +357,28 @@ func (c *Client) InstallTiller() error {
...
+	o := func() error {
+		t, err := c.newTunnel()
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		defer c.closeTunnel(t)
+
+		_, err = c.newHelmClientFromTunnel(t).UpdateRelease(releaseName, path, options...)
+		if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
```
tuommaki: As earlier, `releaseName` to error?

helmclient.go
```diff
@@ -67,13 +68,28 @@ func New(config Config) (*Client, error) {
...
+		_, err = c.newHelmClientFromTunnel(t).DeleteRelease(releaseName, options...)
+		if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+
+		return nil
+	}
+	b := newExponentialBackoff(2*time.Minute, backoff.DefaultMaxInterval)
```
kopiczko: I think it would be good to extract this `2*time.Minute` thing, as this is used in multiple places.

xh3b4sd: How do you want it to look like?

error.go
xh3b4sd: This is about the difference of our custom error to have an error and a matcher and having the string matching magic for the untyped upstream errors. Our error here is correct IMO. The matching error string is defined by upstream and it has to be how it is.

helmclient.go
xh3b4sd: I thought about this but wasn't sure. I can do this though.

xh3b4sd: Done.

helmclient.go
xh3b4sd: Done.

helmclient.go
xh3b4sd: Done.

helmclient.go
xh3b4sd: Done.

helmclient.go
kopiczko: `var backOffMaxWait = 2 * time.Minute`

xh3b4sd: I am going with this now. We can still sort this issue if you like. I am glad to provide a separate PR to improve the situation here. Just give me a shout.

PR #11, opened 2018-04-23
Add InstallTiller to spec.go

rossf7:
Just adds the method to spec.go as I noticed it was missing.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -240,6 +240,8 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


helmclient.go
```diff
@@ -240,6 +240,8 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
 	return nil
 }
 
+// InstallTiller installs Tiller by creating its deployment and waiting for it
+// to start. A service account and cluster role binding are also created.
```
rossf7: Also added GoDoc to add to the existing comments.

PR #12, opened 2018-04-23
Expose PingTiller

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

We need this method exposed so that we know if Tiller is already installed.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -356,6 +356,22 @@ func (c *Client) InstallTiller() error {
 	return nil
 }
 
+// PingTiller checks if Tiller is installed and alive.
+func (c *Client) PingTiller() error {
+	t, err := c.newTunnel()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	defer c.closeTunnel(t)
+
+	err = c.newHelmClientFromTunnel(t).PingTiller()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
 // UpdateReleaseFromTarball updates the given release using the chart packaged
 // in the tarball.
 func (c *Client) UpdateReleaseFromTarball(releaseName, path string, options ...helmclient.UpdateOption) error {
--- a/spec.go
+++ b/spec.go
@@ -27,6 +27,8 @@ type Interface interface {
 	// InstallTiller installs Tiller by creating its deployment and waiting for
 	// it to start. A service account and cluster role binding are also created.
 	InstallTiller() error
+	// PingTiller checks if Tiller is installed and alive.
+	PingTiller() error
 	// UpdateReleaseFromTarball updates the given release using the chart packaged
 	// in the tarball.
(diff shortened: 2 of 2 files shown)


xh3b4sd:
Cann you provide more context? Why can't we just call `InstallTiller` and that is it? It should be idempotent. Alternatively I would like to provide a primitive like `IsTillerInstalled` that clients can use for this purpose. When we want to expose plain helm client features we should just expose the configured client and portforwarding magic.

fgimenez:
Sure, the goal is having a lightweight method for determining if tiller is up. 

In cluster-operator we will need it to bootstrap chart-operator, it would be nice to have a simple way of determining if it should be installed, given that we can only get the helm client from the reconciliation methods (we will need one different tiller for each CR, and the object is not available on resource instantiation, we need to call this https://github.com/giantswarm/cluster-operator/blob/master/pkg/v2/resource/chart/resource.go#L107 when `obj` is available). We can of course call `InstallTiller` every time, but looks like we would be making a lot of unuseful API calls, WDYT?

I like what you propose about `IsTillerInstalled`, will rework this PR in that direction.

xh3b4sd:
Having a `IsTillerInstalled` sounds good.

kopiczko:
> We can of course call InstallTiller every time, but looks like we would be making a lot of unuseful API calls, WDYT?

Unless (as @xh3b4sd proposed already) `InstallTiller` is idempotent and does `PingTiller` check before proceeding with the installation. If the tiller is installed it can return early.

I personally would rename `InstallTiller` to `EnsureTillerInstalled`, make it idempotent and call it like crazy. Naming idea comes from here: https://github.com/giantswarm/operatorkit/blob/1ddbbec19f4b9a9f8f942cb9313fc273e2a6a63c/client/k8scrdclient/k8scrdclient.go#L43

fgimenez:
> Unless (as @xh3b4sd proposed already) InstallTiller is idempotent and does PingTiller check before proceeding with the installation. If the tiller is installed it can return early.

It is already idempotent, the problem I see is that it doesn't return early and causes unneeded API calls in case Tiller is already installed, see https://github.com/giantswarm/helmclient/blob/master/helmclient.go#L245. If we put this method into the reconciliation loop we'll get lots of API calls that could be prevented.

> I personally would rename InstallTiller to EnsureTillerInstalled, make it idempotent and call it like crazy.

Yes, that makes sense, instead of making it idempotent we should make sure it returns early, WDYT?

rossf7:
Yes I like the `EnsureTillerInstalled` name and returning early makes sense.

As well as the case of Tiller already being installed there is also the wait until a new guest cluster has provisioned. In this case I think we need to fail fast if we can't connect to the API server.

kopiczko:
+1 rename and return early.

I think less API calls library user has to make is better. Less internals exposed.

fgimenez:
Ok, closing this one then, thx!

PR #13, opened 2018-04-24
Rename InstallTiller EnsureTillerInstalled and add early check+return

fgimenez:
Also updated interface.


indyfree:
Really nice!

PR #16, opened 2018-04-30
Customizable Tiller namespace

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

Also Tiller namespace is now customizable on client creation. The e2e test is updated to check Tiller installed on `giantswarm` namespace side by side with an already installed `kube-system` one.


spec.go
```diff
@@ -3,11 +3,11 @@ package helmclient
 import "k8s.io/helm/pkg/helm"
 
 const (
-	tillerNamespace     = "kube-system"
-	tillerImageSpec     = "gcr.io/kubernetes-helm/tiller:v2.8.2"
-	tillerLabelSelector = "app=helm,name=tiller"
-	tillerPodName       = "tiller"
-	tillerPort          = 44134
+	tillerDefaultNamespace = "giantswarm"
```
rossf7: I'm not sure about changing the default namespace. We also use helmclient in the host cluster and e2e tests. For e2e tests will the giantswarm namespace always exist?

fgimenez: You are totally right, thanks for pointing it out! Fixed to have `kube-system` as the default PTAL. All the current clients will keep working as until now, and we are able to set the namespace to `giantswarm` when needed.

rossf7: Thanks! Yes I like that we just set the namespace when needed.

PR #18, opened 2018-05-08
Rename tillerPodName to avoid clashes and add logging info

fgimenez:
If we try to deploy tiller on a different namespace the ServiceAccount and ClusterRolebinding might be already there for another tiller deployment, these changes prevent clashes and add logging info.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -138,6 +138,7 @@ func (c *Client) EnsureTillerInstalled() error {
 
 		_, err := c.k8sClient.CoreV1().ServiceAccounts(n).Create(i)
 		if errors.IsAlreadyExists(err) {
+			c.logger.Log("level", "debug", "message", fmt.Sprintf("serviceaccount %s creation failed", tillerPodName), "stack", fmt.Sprintf("%#v", err))
 			// fall through
 		} else if err != nil {
 			return microerror.Mask(err)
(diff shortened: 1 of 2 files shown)


helmclient.go
```diff
@@ -138,6 +138,7 @@ func (c *Client) EnsureTillerInstalled() error {
 
 		_, err := c.k8sClient.CoreV1().ServiceAccounts(n).Create(i)
 		if errors.IsAlreadyExists(err) {
+			c.logger.Log("level", "debug", "message", fmt.Sprintf("serviceaccount %s creation failed", tillerPodName), "stack", fmt.Sprintf("%#v", err))
```
indyfree: Nice to have some logging <3

spec.go
```diff
@@ -6,7 +6,7 @@ const (
 	tillerDefaultNamespace = "kube-system"
 	tillerImageSpec        = "gcr.io/kubernetes-helm/tiller:v2.8.2"
 	tillerLabelSelector    = "app=helm,name=tiller"
-	tillerPodName          = "tiller"
+	tillerPodName          = "tiller-giantswarm"
```
rossf7: I wondered whether the pod name should include the configured namespace. But this makes sense since its our helmclient library.

fgimenez: Exactly, it's in fact not related to the namespace in this case, but we could easily bind the two together if needed.

PR #19, opened 2018-05-09
add additional info for invalid gzip header

fgimenez:
Related to https://github.com/giantswarm/giantswarm/issues/3109

This will give us more insight into the laconic `gzip: invalid header` error.


fgimenez:
> Did it lately pop up again?

Yes, outdated versions of helmclient have started returning it again on `UpdateFunc`, let's see if we can catch the underlying problem. In any case this aims to give more visibility on the possible errors, not just the current ones.

PR #22, opened 2018-05-14
Add function for asserting guest API not available errors

fgimenez:
This will allow us to detect errors related to the guest API not being available and act according to that in consumer code, for instance, we may stop a reconciliation loop if we are getting this kind of error after creating a guest cluster and expect to have the API ready soon.

### DIFF
--- a/error.go
+++ b/error.go
@@ -38,6 +38,7 @@ func IsExecutionFailed(err error) bool {
 	return microerror.Cause(err) == executionFailedError
 }
 
+var guestAPINotAvailableError = microerror.New("Guest API not available")
 var guestNamespaceCreationErrorSuffix = "namespaces/kube-system/serviceaccounts: EOF"
 
 // match example https://play.golang.org/p/ipBkwqlc4Td
@@ -59,6 +60,10 @@ func IsGuestAPINotAvailable(err error) bool {
 		return false
 	}
 	if matched {
+		return true
+	}
+
+	if c == guestAPINotAvailableError {
 		return true
 	}
 
--- a/error.go
+++ b/error.go
@@ -37,6 +37,25 @@ func IsExecutionFailed(err error) bool {
 	return microerror.Cause(err) == executionFailedError
 }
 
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -37,6 +37,25 @@ func IsExecutionFailed(err error) bool {
...
+var guestNamespaceCreationErrorSuffix = "namespaces/kube-system/serviceaccounts: EOF"
+var guestDNSNotReadySuffix = "53: no such host"
+
+// IsGuestAPINotAvailable asserts guestAPINotAvailableError.
+func IsGuestAPINotAvailable(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if strings.HasSuffix(err.Error(), guestNamespaceCreationErrorSuffix) {
```
xh3b4sd: So I like that we work on better error handling and such. I am just wondering if we want to get more type safety if we can. Like the `no such host` errors can be asserted to URL errors or OS errors or something. The other probably to k8s errors. We might get into situations where something else returns `no such host` and then we should maybe get aware of it. WDYT? 

I also do not see the matcher being used.

xh3b4sd: Oh, and I think we should actually only match against the cause. The error message might be difference than the cause's one.

fgimenez: > Like the no such host errors can be asserted to URL errors or OS errors or something. The other probably to k8s errors. We might get into situations where something else returns no such host and then we should maybe get aware of it. WDYT?

Sure, I've updated it to check more precisely an error coming from DNS not available with a regexp, PTAL.

fgimenez: > Oh, and I think we should actually only match against the cause. The error message might be difference than the cause's one.

Sure, done.

fgimenez: > I also do not see the matcher being used.

Yes, this is meant to be used from consumer libraries, IMO having the function in place, and given that it checks for all the possible (attm) error conditions and the custom error it doesn't make too much sense to add an annotation with the custom error, WDYT? External code will just use the function, it won't care if the error carries the custom error message or some other message from external APIs, for instance, the function knows how to handle that.

Anyway, we can discuss this in another place, to align with what we have I've already added the annotation here too.

PR #25, opened 2018-05-22
Add more patterns to detect guest cluster API not ready

fgimenez:
This will identify transient certificate errors like:
```
Post https://api.29bxw.k8s.tarantula.germanycentral.azure.gigantic.io/api/v1/namespaces/giantswarm/serviceaccounts: x509: certificate is valid for ingress.local, not api.29bxw.k8s.tarantula.germanycentral.azure.gigantic.io
```
and
```
Get https://api.29bxw.k8s.tarantula.germanycentral.azure.gigantic.io/api/v1/namespaces/giantswarm: x509: certificate is valid for ingress.local, not api.29bxw.k8s.tarantula.germanycentral.azure.gigantic.io
```
These errors usually happen while the guest API is not fully up, once the API is responding they stop appearing. They will be returned as `guestAPINotAvailableError`, so that consumer libraries can act accordingly (for instance, canceling a reconciliation loop).

### DIFF
--- a/error.go
+++ b/error.go
@@ -9,7 +9,27 @@ import (
 
 const (
 	cannotReuseReleaseErrorPrefix = "cannot re-use"
+	// guestDNSNotReadyPattern is a regular expression representing DNS errors for
+	// the guest API domain.
+	// match example https://play.golang.org/p/ipBkwqlc4Td
+	guestDNSNotReadyPattern = "dial tcp: lookup .* on .*:53: no such host"
+
+	// guestTransientInvalidCertificatePattern regular expression defines the kind
+	// of transient errors related to certificates returned while the guest API is
+	// not fully up.
+	// match example https://play.golang.org/p/iiYvBhPOg4f
+	guestTransientInvalidCertificatePattern = `[Get|Post] https://api\..*: x509: certificate is valid for ingress.local, not api\..*`
 )
+
+var (
+	guestDNSNotReadyRegexp                 *regexp.Regexp
+	guestTransientInvalidCertificateRegexp *regexp.Regexp
+)
+
+func init() {
+	guestDNSNotReadyRegexp = regexp.MustCompile(guestDNSNotReadyPattern)
+	guestTransientInvalidCertificateRegexp = regexp.MustCompile(guestTransientInvalidCertificatePattern)
+}
 
 var cannotReuseReleaseError = microerror.New("cannot reuse release")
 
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -44,6 +44,9 @@ var guestNamespaceCreationErrorSuffix = "namespaces/kube-system/serviceaccounts:
 // match example https://play.golang.org/p/ipBkwqlc4Td
 var guestDNSNotReadyPattern = "dial tcp: lookup .* on .*:53: no such host"
 
+// match example https://play.golang.org/p/iiYvBhPOg4f
```
xh3b4sd: The godoc is not 100% because it should start with the thing it describes by convention.

error.go
```diff
@@ -55,12 +58,19 @@ func IsGuestAPINotAvailable(err error) bool {
...
+		guestDNSNotReadyPattern,
+		guestTransientInvalidCertificatePattern,
 	}
-	if matched {
-		return true
+	for _, pattern := range patterns {
+		matched, matchErr := regexp.MatchString(pattern, c.Error())
+		if matchErr != nil {
+			return false
+		}
```
xh3b4sd: How can `matchErr` occur? Is this maybe just a compilation error? In this case we should compile the regexp on program initialization and just use it more safely here at runtime. That would be super cool.

fgimenez: Good point, done PTAL.

rossf7:
Yes, its the namespace resource that is throwing these errors. `NewK8sClient` will work because this just checks for the secrets and returns a K8s client.

We hit the error the first time we call the API to get the current state.
https://github.com/giantswarm/cluster-operator/blob/aa4012c4974e004b6235e6793c4b7daaab271bbf/pkg/v3/resource/namespace/current.go#L53

Yes calling `helmclient` for this feels strange and incorrect to me. I'll add to [pkgv3/guestcluster/error.go](https://github.com/giantswarm/cluster-operator/blob/aa4012c4974e004b6235e6793c4b7daaab271bbf/pkg/v3/guestcluster/error.go) for now.

If multiple operators need to connect to guest clusters e.g. for event stream we can extract this package to a library. Ideally we'd have this in a single library but I don't see a simple way to achieve that at the moment.

fgimenez:
Ok, makes sense.

fgimenez:
These changes are useful for helmclient by itself, merging.

PR #34, opened 2018-07-26
Add RunReleaseTest to client

rossf7:
See https://github.com/giantswarm/kubernetes-nginx-ingress-controller/pull/26

I've been updating the e2etests to remove uses of `framework.helmCmd` but we use this to run helm tests. So this extends the helmclient to also run tests.


integration/test/basic/basic_test.go
```diff
@@ -51,7 +51,16 @@ func TestInstallChart(t *testing.T) {
...
+	// We need to pass the ValueOverrides option to make the install process
+	// use the default values and prevent errors on nested values.
+	//
+	//     {
+	//      rpc error: code = Unknown desc = render error in "cnr-server-chart/templates/deployment.yaml":
+	//      template: cnr-server-chart/templates/deployment.yaml:20:26:
+	//      executing "cnr-server-chart/templates/deployment.yaml" at <.Values.image.reposi...>: can't evaluate field repository in type interface {}
+	//     }
+	//
+	err = helmClient.InstallFromTarball(tarballPath, "default", helm.ReleaseName(releaseName), helm.ValueOverrides([]byte("{}")))
```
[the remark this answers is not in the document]
indyfree: Agree. Somehow helm handles this shitty

helmclient.go
```diff
@@ -394,6 +395,46 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
+	defer c.closeTunnel(t)
+
+	resChan, errChan := c.newHelmClientFromTunnel(t).RunReleaseTest(releaseName, helmclient.ReleaseTestTimeout(300))
+	if IsReleaseNotFound(err) {
+		return backoff.Permanent(microerror.Maskf(releaseNotFoundError, releaseName))
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for {
```
indyfree: Why do we need the `for` here? In case none of the channels returns something yet?

helmclient.go
```diff
@@ -394,6 +395,46 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
+func (c *Client) RunReleaseTest(releaseName string, options ...helmclient.ReleaseTestOption) error {
+	c.logger.Log("level", "debug", "message", fmt.Sprintf("running tests for release '%s'", releaseName))
+
+	t, err := c.newTunnel()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	defer c.closeTunnel(t)
+
+	resChan, errChan := c.newHelmClientFromTunnel(t).RunReleaseTest(releaseName, helmclient.ReleaseTestTimeout(300))
```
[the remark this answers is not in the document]
indyfree: Pretty sweet to use the helm RunReleaseTest here <3

integration/test/basic/fixtures/test-chart/templates/test/tests.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ .Values.pod.name }}-tests
+  namespace: {{ .Values.namespace }}
+data:
+  run.sh: |-
+    @test "Testing nginx is reachable" {
+      response=$(curl -I {{ .Values.service.name }}.{{ .Values.namespace }} 2> /dev/null | head -n 1 | cut -d$' ' -f2)
+      [ "$response" -eq 200 ]
+    }
```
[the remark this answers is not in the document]
indyfree: <3

spec.go
```diff
@@ -29,6 +29,8 @@ type Interface interface {
 	GetReleaseHistory(releaseName string) (*ReleaseHistory, error)
 	// InstallFromTarball installs a Helm Chart packaged in the given tarball.
 	InstallFromTarball(path, ns string, options ...helm.InstallOption) error
+	// RunReleaseTest runs the tests for a Helm Release.
+	RunReleaseTest(releaseName string, options ...helm.ReleaseTestOption) error
```
indyfree: Don't know if it was just me. But maybe make it a bit clearer in the godoc that we're actually using the helmclients 'RunReleaseTest' (equivalent to `> helm test` command) method here, and not just do some testing the release on our own?

indyfree: I think it was just me and I didn't read carefully enough

helmclient.go
rossf7: Helm sends multiple results to the resChan. The for loop is so we capture all of them. The time.After timeout is to make sure we don't get stuck in the loop.

integration/test/basic/basic_test.go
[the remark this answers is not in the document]
rossf7: Maybe helmclient could help here and pass the empty values if none are set? Not sure.

spec.go
kopiczko: I think putting it does similar job to the `helm test` wouldn't hurt.

integration/test/basic/basic_test.go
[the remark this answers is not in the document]
rossf7: Cool I think our client can help there. Its a shame the upstream client doesn't handle it better.

I've created https://github.com/giantswarm/giantswarm/issues/3855 for this.

spec.go
rossf7: Good idea. I've updated the godoc.

helmclient.go
```diff
@@ -394,6 +395,47 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
+	if IsReleaseNotFound(err) {
+		return backoff.Permanent(microerror.Maskf(releaseNotFoundError, releaseName))
```
MarcelMue: What does `backoff.Permanent` do here?

helmclient.go
```diff
@@ -394,6 +395,47 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
 	return nil
 }
 
+// RunReleaseTest runs the tests for a Helm Release. The releaseName is the
+// name of the Helm Release that is set when the Helm Chart is installed. This
+// is the same action as running the helm test command.
+func (c *Client) RunReleaseTest(releaseName string, options ...helmclient.ReleaseTestOption) error {
+	c.logger.Log("level", "debug", "message", fmt.Sprintf("running tests for release '%s'", releaseName))
```
MarcelMue: Little nit: `running release tests for ...` seems more intuitive to me. Not sure though, I feel the naming in this whole helm/appr world is not very clear to me.

helmclient.go
```diff
@@ -394,6 +395,46 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
+			c.logger.Log("level", "debug", "message", res.Msg)
+
+			switch res.Status {
+			case hapirelease.TestRun_SUCCESS:
+				c.logger.Log("level", "debug", "message", fmt.Sprintf("successfully ran tests for release '%s'", releaseName))
+				return nil
+			case hapirelease.TestRun_FAILURE:
+				return microerror.Maskf(testReleaseFailureError, "'%s' has failed tests", releaseName)
+			}
+		case <-time.After(300 * time.Second):
```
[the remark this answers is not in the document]
MarcelMue: Could we put this in a `const`? I think we have a bad habit of having our timeouts/waits/backoffs in magic numbers spread around our code way too much.

helmclient.go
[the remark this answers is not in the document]
MarcelMue: Same as with the timeout below.

helmclient.go
rossf7: Godoc says it wraps the error in a permanent error. It's used in all the other methods so I used it for consistency.

helmclient.go
rossf7: Thanks that's nicer I'll update. Yes helm and appr both having releases is a pain. Especially as it comes out of essentially the same community.

PR #37, opened 2018-08-08
Use giantswarm/backoff

rossf7:
Just updating to use the new library.


integration/test/basic/main_test.go
```diff
@@ -31,8 +31,9 @@ func init() {
 		c := framework.HostConfig{
 			Logger: logger,
 
-			ClusterID:  "someval",
-			VaultToken: "someval",
+			ClusterID:       "someval",
+			TargetNamespace: "default",
```
rossf7: Setting the namespace due to the e2e-harness change.

PR #38, opened 2018-08-09
Remove setting target namespace

rossf7:
Not needed due to https://github.com/giantswarm/e2e-harness/pull/114 ❤️


rossf7:
OK works better now I actually updated vendor!

PR #39, opened 2018-08-20
Expose PingTiller

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/3925

Makes available PingTiller to external consumers of the library, this will be useful to create an alert on Tiller unreachable.


fgimenez:
Good catch! Added to the interface too.

PR #41, opened 2018-08-29
Set helm init's max history param

fgimenez:
Aims to prevent an unlimited number of previous release versions to be stored. See https://gigantic.slack.com/archives/C02MS174K/p1535545641000100

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -192,6 +192,7 @@ func (c *Client) EnsureTillerInstalled() error {
 	{
 		o := &installer.Options{
 			ImageSpec:      tillerImageSpec,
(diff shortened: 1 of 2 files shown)


pipo02mix:
Does it affect existing tiller installations?

fgimenez:
> Does it affect existing tiller installations?

I don't think so, this will only be set while initializing tiller.

> Also at some point we'll need to be able to upgrade the tiller pod version. Should we have an issue for that?

Indeed, we should update the client dependencies too when we bump tiller version, good to have an issue for tracking.

pipo02mix:
This helps with https://github.com/giantswarm/giantswarm/issues/4064

PR #44, opened 2018-09-18
return tillerNotFound error

kopiczko:
... instead of meaningless pod not found.

Towards https://github.com/giantswarm/giantswarm/issues/4059.

Again this allows to build idempotent code around that function call.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -543,7 +543,9 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 	}
 
 	podName, err := getPodName(c.k8sClient, tillerLabelSelector, c.tillerNamespace)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


helmclient.go
```diff
@@ -583,7 +585,7 @@ func getPodName(client kubernetes.Interface, labelSelector, namespace string) (s
 		return "", microerror.Maskf(tooManyResultsError, "%d", len(pods.Items))
 	}
 	if len(pods.Items) == 0 {
-		return "", microerror.Maskf(podNotFoundError, "%s", labelSelector)
+		return "", microerror.Maskf(notFoundError, "%s", labelSelector)
```
kopiczko: `tooManyResultsError` error was generic one. This was specific. I think this method is so well scoped that returning generic error makes more sense here.

PR #45, opened 2018-09-19
fix tiller not retry found handling

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4059

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -93,7 +93,7 @@ func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteO
 
 		_, err = c.newHelmClientFromTunnel(t).DeleteRelease(releaseName, options...)
 		if IsReleaseNotFound(err) {
-			return backoff.Permanent(microerror.Maskf(releaseNotFoundError, releaseName))
+			return backoff.Permanent(microerror.Mask(err))
 		} else if IsTillerNotFound(err) {
 			return backoff.Permanent(microerror.Mask(err))
 		} else if err != nil {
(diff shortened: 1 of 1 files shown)


helmclient.go
```diff
@@ -382,10 +388,12 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
 		defer c.closeTunnel(t)
 
 		release, err := c.newHelmClientFromTunnel(t).InstallRelease(path, ns, options...)
-		if IsReleaseNotFound(err) {
-			return backoff.Permanent(releaseNotFoundError)
-		} else if IsCannotReuseRelease(err) {
+		if IsCannotReuseRelease(err) {
 			return backoff.Permanent(cannotReuseReleaseError)
+		} else if IsReleaseNotFound(err) {
+			return backoff.Permanent(releaseNotFoundError)
```
xh3b4sd: Any reason why some errors are masked and some aren't?

kopiczko: This is just moved down so the order is alphabetical. I didn't change anything so don't ask me for reasons :)

kopiczko: PTALA

xh3b4sd: So why did you change all the errors now? Feels like this will now just break left and right. Are you confident with this? Because, I am not.

kopiczko: What can be broken? All errors are from the same package.

helmclient.go
```diff
@@ -382,10 +388,12 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
-		if IsReleaseNotFound(err) {
-			return backoff.Permanent(releaseNotFoundError)
-		} else if IsCannotReuseRelease(err) {
-			return backoff.Permanent(cannotReuseReleaseError)
+		if IsCannotReuseRelease(err) {
+			return backoff.Permanent(microerror.Mask(err))
+		} else if IsReleaseNotFound(err) {
+			return backoff.Permanent(microerror.Mask(err))
+		} else if IsTillerNotFound(err) {
+			return backoff.Permanent(microerror.Mask(err))
```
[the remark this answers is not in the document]
kopiczko: I like that. However I remember Mr. @xh3b4sd preferring `if` statements.

kopiczko:
I need to have this bug fixed. This is blocking me. Please approve or tell what needs to be changed in order to get an approval.

helmclient.go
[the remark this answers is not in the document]
kopiczko: This is not going to work https://play.golang.org/p/_PgTTBw_lRQ

PR #47, opened 2018-09-27
Ping helm version to 2.8.2

tuommaki:
`=` prefix was missing from Helm version pinning. Added it and
"downgraded" Helm to correct version.


Gopkg.toml
```diff
@@ -84,7 +84,7 @@ required = [
 
 [[constraint]]
   name = "k8s.io/helm"
-  version = "2.8.2"
+  version = "=2.8.2"
```
tuommaki: https://golang.github.io/dep/docs/Gopkg.toml.html#version-rules

PR #49, opened 2018-10-19
integration: improve setup

kopiczko:


integration/setup/setup.go
```diff
@@ -3,33 +3,41 @@
...
+	if err != nil {
+		config.Logger.LogCtx(ctx, "level", "error", "message", "", "stack", fmt.Sprintf("%#v", err))
+		os.Exit(1)
 	}
 
-	if v == 0 {
-		m.Run()
-	}
+	os.Exit(exitCode)
+}
```
kopiczko: Diff is ugly. This is just:

```go
func Setup(m *testing.M, config Config) {
	ctx := context.Background()

	exitCode, err := setup(ctx, m, config)
	if err != nil {
		config.Logger.LogCtx(ctx, "level", "error", "message", "", "stack", fmt.Sprintf("%#v", err))
		os.Exit(1)
	}

	os.Exit(exitCode)
}
```

integration/setup/setup.go
```diff
@@ -3,33 +3,41 @@
...
-		// only do full teardown when not on CI
-		if os.Getenv("CIRCLECI") != "true" {
-			// TODO there should be error handling for the framework teardown.
-			h.Teardown()
+func setup(ctx context.Context, m *testing.M, config Config) (int, error) {
+	var err error
+	teardown := !env.CircleCI() && !env.KeepResources()
+
+	{
+		// TODO this should be named EnsureNamespaceCreated
```
kopiczko: I'm going to address all todos in subsequent PRs. I won't create issues.

integration/setup/setup.go
```diff
@@ -3,33 +3,41 @@
...
+		err = config.K8sSetup.EnsureNamespace(ctx, tillerNamespace)
+		if err != nil {
+			return 1, microerror.Mask(err)
+		}
+		if teardown {
+			// TODO defer config.K8sSetup.EnsureNamespaceDeleted(ctx, tillerNamespace)
```
kopiczko: I was talking about that already. `defer` instead of `v == 1`. Deferred functions are pushed to stack so we can be sure deletion will happen in desired order.

integration/test/basic/basic_test.go
```diff
@@ -6,46 +6,18 @@ import (
...
-		Logger:          l,
-		K8sClient:       cs,
-		RestConfig:      config,
-		TillerNamespace: "giantswarm",
-	}
-
-	helmClient, err := helmclient.New(c)
-	if err != nil {
-		t.Fatalf("could not create helm client %v", err)
-	}
```
kopiczko: I want more tests here. So I'd like to simplify setup.

I want a test for IsReleaseAlreadyCreated (not introduced yet).
My goal is to make release installation idempotent in e2e-harness. We are refactoring things. Let's not take more debt and do things right.

PR #50, opened 2018-10-19
integration, vendor: delete tiller namespace after e2e tests

kopiczko:


integration/setup/setup.go
```diff
@@ -29,13 +29,17 @@ func setup(ctx context.Context, m *testing.M, config Config) (int, error) {
 	teardown := !env.CircleCI() && !env.KeepResources()
 
 	{
-		// TODO this should be named EnsureNamespaceCreated
 		err = config.K8sSetup.EnsureNamespace(ctx, tillerNamespace)
```
xh3b4sd: Well this name should be changed now.

PR #53, opened 2018-10-23
helmclient, integration: improve error handling for InstallRelease

kopiczko:


integration/test/basic/fixtures/tb-chart/templates/test/test-runner.yaml
```diff
@@ -1,41 +0,0 @@
-apiVersion: v1
```
kopiczko: This `tb-chart` is leftover from previous PR. In this PR I created another test-chart which is basically as other one, but w/o tests. So the diff looks like that.

helmclient.go
```diff
@@ -394,7 +394,16 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
...
+		} else if IsReleaseAlreadyExists(err) {
+			return backoff.Permanent(microerror.Mask(err))
+		} else if IsTarballNotFound(err) {
+			return backoff.Permanent(microerror.Mask(err))
 		} else if IsReleaseNotFound(err) {
+			// This is out of alphabetical order. But
+			// IsReleaseNotFound is rather bold And I want to catch
+			// other errors first. I don't know why do we even
+			// match by it here. Also it would be good to have
+			// tests of all the matchers.
```
kopiczko: As I wrote here. I can't tell how this error can happen during install. Also matching isn't very accurate IMO.
If there was a case for this error I would write a test.

helmclient.go
```diff
@@ -394,7 +394,16 @@ func (c *Client) InstallFromTarball(path, ns string, options ...helmclient.Insta
 		release, err := c.newHelmClientFromTunnel(t).InstallRelease(path, ns, options...)
 		if IsCannotReuseRelease(err) {
```
kopiczko: I would also write a test for that matches but I don't understand when this happens.

helmclient.go
xh3b4sd: The method is called `InstallRelease`. So I assume when you try to install a release that does not exist an "release not found" error is returned. That sounds pretty legit to me

kopiczko: The release must not exist in order to install it. Otherwise `IsReleaseAlreadyExists` I just introduced is returned.
If you speak about the chart tarball then I added `IsTarballNotFound`.

Still I don't get how it can return `IsReleaseNotFound`.

kopiczko: Also a test for `IsTarballNotFound` is written below.

xh3b4sd: Oh wow, yeah. Totally. Maybe that was just bad copy/paste. Lets drop it then, because this does not make any sense and when the installation is not returning such a specific error then this is useless.

kopiczko: I forgot to remove. https://github.com/giantswarm/helmclient/pull/54.

PR #55, opened 2018-10-23
integraton/test/deleterelease: add test for IsReleaseNotFound

kopiczko:


integration/test/deleterelease/fixtures/test-chart/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
kopiczko: All below (fixtures dir) is copy/paste.

PR #56, opened 2018-10-24
cricleci, integration: fix & run deleterelease test

kopiczko:


.circleci/config.yml
```diff
@@ -69,7 +74,9 @@ workflows:
       - e2eTestBasic:
           requires:
           - build
-      - build
```
kopiczko: Bad copy/pasting.

.circleci/config.yml
```diff
@@ -69,7 +74,9 @@ workflows:
...
       - e2eTestInstallRelease:
           requires:
           - build
+      - e2eTestDeleteRelease:
+          requires:
+          - build
```
kopiczko: The test wasn't running at all.

PR #58, opened 2018-11-16
add ctx; improve logs; update backoff code

kopiczko:
- Add ctx parameter to the Methods
- Rename InstallFromTarball to InstallReleaseFromTarball
- Use LogCtx instead of Log
- Use latest backoff advancements
- Improve logging in general
- Add TODO about role bindings

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -83,7 +83,7 @@ func New(config Config) (*Client, error) {
 }
 
 // DeleteRelease uninstalls a chart given its release name.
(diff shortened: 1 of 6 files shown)


helmclient.go
```diff
@@ -381,15 +400,15 @@ func (c *Client) GetReleaseHistory(releaseName string) (*ReleaseHistory, error)
 }
 
 // InstallFromTarball installs a chart packaged in the given tarball.
```
indyfree:
```suggestion
// InstallReleaseFromTarball installs a chart packaged in the given tarball.
```

PR #61, opened 2018-11-28
vendor: update

kopiczko:


kopiczko:
@MarcelMue looks like https://github.com/giantswarm/giantswarm/issues/4721
We didn't have the logs but with https://github.com/giantswarm/micrologger/pull/19 merged we should have them.

PR #64, opened 2018-11-28
aligns badge with other projects

xh3b4sd:


xh3b4sd:
The last merge to master broke for some reason. See https://circleci.com/gh/giantswarm/helmclient/478.

xh3b4sd:
The basic test hangs on teardown like that. Do you know why that is? 

screen shot 2018-11-28 at 17 44 07

kopiczko:
I don't. The code suggests there are couple bash commands to be executed. For w/e reason it hangs. This isn't critical to even run this inside the CI. And it should be translated to hlemclient/client-go calls. So I won't dig deeper into it now.

xh3b4sd:
Do I have to fix this now? Do we rack this somewhere? What should we do about it? The CI fails after all.

kopiczko:
If it happens again and becomes annoying it should be fixed eventually. I know it isn't a great answer but we have so many more urgent things that I don't want to distract myself with that in scope of flaky e2e. If it happens often, then we will track that and fix. This is the first time I see it. So I'd say just merge. Or rerun the test if you want to see everything green.

xh3b4sd:
Then please track that somewhere. I go with this now.

PR #65, opened 2018-11-29
integration, vendor: remove e2esetup dependency from code (#63)

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4709.

We were using e2esetup to create/delete tiller namespace in tests. We
don't want the dependency in the code though because e2esetup depends on
helmclient. This creates potential circular dependency problem and makes
vendoring hard.


integration/setup/k8s/setup.go
```diff
@@ -13,6 +13,14 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+/*
+	NOTE: This code is copied from e2esetup/k8s package to avoid circular
+	dependencies. All changes made there should be reflected in e2esetup.
+
+		See https://github.com/giantswarm/e2esetup/pull/13
+
+*/
```
kopiczko: The diff is small because git thinks the file was moved from vendor directory. This is a full copy.

xh3b4sd:
Why are we doing this? I understand this for the single constant in `e2e-harness` but this here goes a bit too far IMO. Why don't we have a common lib? Let's please not copy so much and try to "fix" it with a comment.

kopiczko:
> Why are we doing this?

Circular dependencies. I explained that in the description.

> Why don't we have a common lib?

Maybe at some point that would be a good idea to have it. I'm not sure how should it look like ATM and what would be its scope. Things are changing in e2e and I don't want to constrain myself. This change debatable I agree. I'll try to do the upgrades with overrides.

xh3b4sd:
If you like we can just hang and come up with some proper dependency graph. Or just create a common lib for that.

xh3b4sd:
Cool. Just give me a shout.

kopiczko:
I chatted with @xh3b4sd. We agreed it's better to just use `[[override]]` for `k8s.io` libs in `helmclient`. We are not sure what is the future of our testing libraries. We should figure that out first before extracting some random stuff (which may require e2e tests itself) to a separate repo.

PR #67, opened 2018-12-05
Add getTillerImage

rossf7:

This is prep for updating the tiller pod. Adds `getTillerImage` which will be used to check if the image is outdated.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -634,7 +634,7 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 		return nil, nil
 	}
 
-	podName, err := getPodName(c.k8sClient, tillerLabelSelector, c.tillerNamespace)
+	pod, err := getPod(c.k8sClient, tillerLabelSelector, c.tillerNamespace)
 	if IsNotFound(err) {
 		return nil, microerror.Maskf(tillerNotFoundError, "label selector: %#q namespace: %#q", tillerLabelSelector, c.tillerNamespace)
 	} else if err != nil {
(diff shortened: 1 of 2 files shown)


helmclient.go
```diff
@@ -664,26 +664,55 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 	return tunnel, nil
 }
 
-func getPodName(client kubernetes.Interface, labelSelector, namespace string) (string, error) {
+func getPod(client kubernetes.Interface, labelSelector, namespace string) (*corev1.Pod, error) {
```
rossf7: getPod is used to get the name and the image.

helmclient.go
```diff
@@ -664,26 +664,55 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
...
 	}
 	if len(pods.Items) == 0 {
-		return "", microerror.Maskf(notFoundError, "%s", labelSelector)
+		return nil, microerror.Maskf(notFoundError, "%s", labelSelector)
+	}
+
+	return &pods.Items[0], nil
+}
+
+func getPodName(client kubernetes.Interface, labelSelector, namespace string) (string, error) {
```
rossf7: Wrapper for the current function.

helmclient.go
```diff
@@ -664,26 +664,55 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
...
+	pod, err := getPod(client, labelSelector, namespace)
+	if err != nil {
+		return "", microerror.Mask(err)
 	}
-	pod := pods.Items[0]
 
 	return pod.Name, nil
 }
 
+func getTillerImage(client kubernetes.Interface, labelSelector, namespace string) (string, error) {
```
rossf7: New function for use in later PRs.

helmclient.go
```diff
@@ -664,24 +664,39 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
...
+
+	return &pods.Items[0], nil
+}
+
+func getTillerImage(pod *corev1.Pod) (string, error) {
+	if len(pod.Spec.Containers) > 1 {
+		return "", microerror.Maskf(tooManyResultsError, "%d", len(pod.Spec.Containers))
+	}
+	if len(pod.Spec.Containers) == 0 {
+		return "", microerror.Maskf(notFoundError, "%d", len(pod.Spec.Containers))
```
MarcelMue: This does not look helpful, this would just print `not found 0` or something like that. I think leaving the number out completely would be just fine here.

PR #68, opened 2018-12-05
Add check for tiller pod being outdated

rossf7:

Adds check for whether the tiller pod is outdated. This will be used in a later PR to trigger the upgrade process.

### DIFF
--- a/error.go
+++ b/error.go
@@ -186,15 +186,6 @@ func IsTestReleaseTimeout(err error) bool {
 	return microerror.Cause(err) == testReleaseTimeoutError
 }
 
-var tillerImageInvalidError = &microerror.Error{
-	Kind: "tillerImageInvalidError",
-}
-
-// IsTillerImageInvalid asserts tillerImageInvalidError.
-func IsTillerImageInvalid(err error) bool {
-	return microerror.Cause(err) == tillerImageInvalidError
-}
-
 var tillerInstallationFailedError = &microerror.Error{
 	Kind: "tillerInstallationFailedError",
 }
--- a/error.go
+++ b/error.go
@@ -186,13 +186,13 @@ func IsTestReleaseTimeout(err error) bool {
 	return microerror.Cause(err) == testReleaseTimeoutError
 }
 
-var tillerImageNotFoundError = &microerror.Error{
-	Kind: "tillerImageNotFoundError",
+var tillerImageInvalidError = &microerror.Error{
+	Kind: "tillerImageInvalidError",
 }
 
-// IsTillerImageNotFound asserts tillerImageNotFoundError.
-func IsTillerImageNotFound(err error) bool {
-	return microerror.Cause(err) == tillerImageNotFoundError
+// IsTillerImageInvalid asserts tillerImageInvalidError.
+func IsTillerImageInvalid(err error) bool {
+	return microerror.Cause(err) == tillerImageInvalidError
 }
 
 var tillerInstallationFailedError = &microerror.Error{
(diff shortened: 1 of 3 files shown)


helmclient.go
```diff
@@ -664,6 +666,31 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 	return tunnel, nil
 }
 
+// filterList returns a list scrubbed of old releases.
+// See https://github.com/helm/helm/blob/3a8a797eab0e1d02456c7944bf41631546ee2e47/cmd/helm/list.go#L197.
```
rossf7: Just tidying up as this was in the wrong place.

error.go
```diff
@@ -186,13 +186,13 @@ func IsTestReleaseTimeout(err error) bool {
 	return microerror.Cause(err) == testReleaseTimeoutError
 }
 
-var tillerImageNotFoundError = &microerror.Error{
-	Kind: "tillerImageNotFoundError",
+var tillerImageInvalidError = &microerror.Error{
+	Kind: "tillerImageInvalidError",
```
rossf7: Renamed the error to make it more generic. It is returned if we can't get the image or if the format is invalid.

helmclient.go
```diff
@@ -693,12 +720,68 @@ func getTillerImage(pod *corev1.Pod) (string, error) {
...
+
+	desiredTillerVersion, err := parseTillerVersion(tillerImageSpec)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for i, currentVersion := range currentTillerVersion {
+		desiredVersion := desiredTillerVersion[i]
+		if currentVersion < desiredVersion {
+			return microerror.Maskf(tillerOutdatedError, "%#q older than %#q", currentTillerVersion, desiredTillerVersion)
```
rossf7: Error is only returned if the version is older. As there may be components running different versions of helmclient and we don't want to downgrade the version.

error.go
kopiczko: I don't think creating very specific errors just in case makes sense in general. We don't track where we handle them and then are afraid to change them. I'd go with `executionFailedError` with a nice description everywhere and extract a specific error if it needs to be handled somewhere.

kopiczko: My comment applies also to `tillerImageInvalidError` and other newly intorduced errors.

helmclient.go
```diff
@@ -693,12 +720,68 @@ func getTillerImage(pod *corev1.Pod) (string, error) {
...
+
+	versionParts := strings.Split(tagParts[0], ".")
+	if len(versionParts) != 3 {
+		return version, microerror.Maskf(tillerImageInvalidError, "%s", tillerImage)
+	}
+
+	for i, s := range versionParts {
+		v, err := strconv.Atoi(s)
+		if err != nil {
+			return version, microerror.Mask(err)
```
kopiczko: IMO ere it would be good to have an error saying something like `"can not convert %d part of version %#q", i, tag`.

helmclient.go
```diff
@@ -693,12 +720,68 @@ func getTillerImage(pod *corev1.Pod) (string, error) {
...
 	tillerImage := pod.Spec.Containers[0].Image
 	if tillerImage == "" {
-		return "", microerror.Mask(tillerImageNotFoundError)
+		return "", microerror.Mask(tillerImageInvalidError)
 	}
 
 	return tillerImage, nil
 }
 
+func isTillerOutdated(pod *corev1.Pod) error {
```
tomahawk28: What about extending return type to `(bool, error)`? currently we are trying to use `error` as response and error check as well.

helmclient.go
kopiczko: Should we still do requests against newer version?

error.go
rossf7: OK cool that makes sense. I see the problem with making it hard to change errors later. 

I'm going to keep `tillerOutdatedError` because I'll need to test for that.

helmclient.go
rossf7: @tomahawk28 Generally we aim for fewer return params and we have matchers for our errors. So I'll use the error matcher to test for this error specifically.

https://github.com/giantswarm/helmclient/blob/f6ff1a2c6194749573ab2db9185bcc968614d6eb/error.go#L212

Our error handling is part of `microkit` and we have docs in `fmt`. Good to read if you haven't seen them already. https://github.com/giantswarm/fmt/blob/master/go/errors.md

helmclient.go
rossf7: My concern is on rolling out the updated `helmclient` version. If we block requests on newer versions too then for example we need to deploy chart-operator and cluster-operator at the same time.

But maybe that is necessary. Not certain yet. I'll keep it in mind and include in my testing.

PR #70, opened 2018-12-10
Add structure for upgrade tiller e2e test

rossf7:

Adds structure for new e2e test for the upgrade process.

### DIFF
--- a/integration/setup/config.go
+++ b/integration/setup/config.go
@@ -18,6 +18,7 @@ const (
 
 type Config struct {
 	HelmClient *helmclient.Client
+	K8sClient  kubernetes.Interface
 	K8sSetup   *k8s.Setup
 	Logger     micrologger.Logger
 }
@@ -82,6 +83,7 @@ func NewConfig() (Config, error) {
 
 	c := Config{
 		HelmClient: helmClient,
+		K8sClient:  k8sClient,
 		K8sSetup:   k8sSetup,
 		Logger:     logger,
 	}
--- a/integration/setup/config.go
+++ b/integration/setup/config.go
@@ -3,7 +3,6 @@
 package setup
 
 import (
-	"github.com/giantswarm/e2esetup/k8s"
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
@@ -19,7 +18,6 @@ const (
 type Config struct {
 	HelmClient *helmclient.Client
 	K8sClient  kubernetes.Interface
-	K8sSetup   *k8s.Setup
 	Logger     micrologger.Logger
 }
 
@@ -52,19 +50,6 @@ func NewConfig() (Config, error) {
 		}
 	}
 
-	var k8sSetup *k8s.Setup
-	{
-		c := k8s.SetupConfig{
-			K8sClient: k8sClient,
-			Logger:    logger,
-		}
-
-		k8sSetup, err = k8s.NewSetup(c)
-		if err != nil {
-			return Config{}, microerror.Mask(err)
-		}
-	}
-
 	var helmClient *helmclient.Client
 	{
 		c := helmclient.Config{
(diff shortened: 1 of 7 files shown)


integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -0,0 +1,109 @@
...
+				return microerror.Maskf(tooManyResultsError, "%d", len(l.Items))
+			}
+			if len(l.Items) == 0 {
+				return microerror.Maskf(notFoundError, "%s", labelSelector)
+			}
+
+			d = &l.Items[0]
+			if d.Status.AvailableReplicas != 1 && d.Status.ReadyReplicas != 1 {
+				return microerror.Maskf(notFoundError, "tiller deployment not ready expected 1 pod found %d available %d ready", d.Status.AvailableReplicas, d.Status.ReadyReplicas)
+			}
```
rossf7: Wait for upgrade to complete before returning the deployment.

integration/setup/config.go
```diff
@@ -18,6 +18,7 @@ const (
 
 type Config struct {
 	HelmClient *helmclient.Client
+	K8sClient  kubernetes.Interface
```
rossf7: K8sClient is used to get the tiller image from the deployment.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	tillerImage := d.Spec.Template.Spec.Containers[0].Image
+	if tillerImage == "" {
+		return "", microerror.Maskf(notFoundError, "tiller image is empty")
+	}
+
+	return tillerImage, nil
+}
+
+func updateTillerImage(ctx context.Context, namespace, labelSelector, tillerImage string) error {
```
rossf7: In next PR this will update the tiller image to a previous version to test the upgrade.

integration/setup/config.go
kopiczko: What do you think about setting up `K8sSetup` below as an internal? We have all dependencies in place and the change is breaking anyway?

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func getTillerImage(ctx context.Context, namespace, labelSelector string) (string, error) {
+	d, err := getTillerDeployment(ctx, namespace, labelSelector)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	if len(d.Spec.Template.Spec.Containers) > 1 {
+		return "", microerror.Maskf(tooManyResultsError, "%d", len(d.Spec.Template.Spec.Containers))
```
kopiczko: Similar to our previous discussion. Those errors aren't really handled. They are just passing message.
I suggest replacing that with something like:
```go
 	if len(d.Spec.Template.Spec.Containers) != 1 {
		return "", microerror.Maskf(executionFailedError, "Spec.Template.Spec.Containers == %d, want 1", len(d.Spec.Template.Spec.Containers))
	}
```

integration/test/upgradetiller/upgrade_tiller_test.go
kopiczko: I prefer that to be introduced with the next PR in that case so we can see it as a whole. It's easy to forget what `updateTillerImage` does when you are out of context.

integration/setup/config.go
rossf7: Yes good point. Done.

integration/test/upgradetiller/upgrade_tiller_test.go
rossf7: Good call. Done.

integration/test/upgradetiller/upgrade_tiller_test.go
rossf7: Still getting used to error handling change but makes total sense. Done.

PR #71, opened 2018-12-10
Upgrade tiller when pod is outdated

rossf7:
Towards giantswarm/giantswarm#4809

This PR extends the client so it can upgrade the tiller pod. Final PR will have the upgrade to helm `2.12.0` providing it passes testing.

One important thing to note is I've switched to using the [Masterminds/semver](https://github.com/Masterminds/semver) library. When I dug deeper I found this is used by upstream. I think its more reliable to use the same library.

https://github.com/helm/helm/blob/v2.12.0/cmd/helm/installer/install.go#L86-L104


helmclient.go
```diff
@@ -228,18 +228,55 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		}
 	}
 
-	// Install the tiller deployment in the tenant cluster.
-	{
-		c.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating tiller in namespace %#q", c.tillerNamespace))
+	var err error
+	var installTiller bool
+	var pod *corev1.Pod
+	var upgradeTiller bool
```
rossf7: We get the tiller pod and check if its outdated.

helmclient.go
```diff
@@ -259,6 +296,28 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		if err != nil {
 			return microerror.Mask(err)
 		}
+	} else if !installTiller && upgradeTiller {
```
rossf7: This is the new code for the upgrade.

helmclient.go
```diff
@@ -273,6 +332,8 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 			t, err := c.newTunnel()
 			if IsTillerNotFound(err) {
 				return backoff.Permanent(microerror.Mask(err))
+			} else if upgradeTiller && IsTooManyResults(err) {
+				return microerror.Maskf(err, "too many tiller pods due to upgrade")
```
rossf7: This is to make it clear in the logs that an upgrade is happening.

helmclient.go
```diff
@@ -446,6 +507,7 @@ func (c *Client) InstallReleaseFromTarball(ctx context.Context, path, ns string,
 	return nil
 }
 
+// ListReleaseContents gets the current status of all Helm Releases.
```
rossf7: Just tidying up.

helmclient.go
```diff
@@ -742,41 +804,26 @@ func isTillerOutdated(pod *corev1.Pod) error {
 		return microerror.Mask(err)
 	}
 
-	for i, currentVersion := range currentTillerVersion {
-		desiredVersion := desiredTillerVersion[i]
-		if currentVersion < desiredVersion {
-			return microerror.Maskf(tillerOutdatedError, "%#q older than %#q", currentTillerImage, tillerImageSpec)
-		}
+	if currentTillerVersion.LessThan(desiredTillerVersion) {
```
rossf7: Comparison is done by the library.

helmclient.go
```diff
@@ -742,41 +804,26 @@ func isTillerOutdated(pod *corev1.Pod) error {
...
 	return nil
 }
 
-func parseTillerVersion(tillerImage string) ([]int, error) {
-	version := make([]int, 3)
-
+func parseTillerVersion(tillerImage string) (*semver.Version, error) {
 	// Tiller image tag has the version.
-	imageParts := strings.Split(tillerImage, ":v")
+	imageParts := strings.Split(tillerImage, ":")
```
rossf7: Library removes the `v` prefix which is kinda nice.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -26,11 +26,41 @@ func TestUpgradeTiller(t *testing.T) {
 
 	namespace := "giantswarm"
 	labelSelector := "app=helm,name=tiller"
+	outdatedTillerImage := "gcr.io/kubernetes-helm:v2.7.2"
```
rossf7: In the final PR I'll put this back to our image and `2.8.2 `.

Its worth noting the upgrade will only succeed if the tiller version is different. We have `2.8.1` and `2.8.2` tagged but they have the same tiller version.

Gopkg.toml
```diff
@@ -90,6 +90,10 @@ required = [
   name = "github.com/mholt/archiver"
   version = "=2.1.0"
 
+[[constraint]]
+  name = "github.com/Masterminds/semver"
```
MarcelMue: We usually use `"github.com/coreos/go-semver/semver"` - not a big deal but maybe keeping it aligned makes sense?

MarcelMue: But I see that we apparently use this semver package here. Not sure if we want to create a boyscouting issue to align that.

helmclient.go
MarcelMue: So we would check for the `TooManyResults` error when we call `EnsureTillerInstalled`?
I would vouch for no error at all and waiting instead. That would match the behaviour described in the godoc of this function IMO.

helmclient.go
```diff
@@ -228,18 +228,55 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
...
+
+		err := backoff.RetryNotify(o, b, n)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	if pod != nil {
+		err = isTillerOutdated(pod)
+		if IsTillerOutdated(err) {
```
pipo02mix: For me, this is a bit confusing, maybe because of my background. But why instead of return a bool (and error if it is needed), we return an error when tiller is outdated?

helmclient.go
pipo02mix: Just wondering if it would not be better to split this big function in small chunks (moving out the upgrade, install,... parts) and improve readability

Gopkg.toml
rossf7: > We usually use "github.com/coreos/go-semver/semver" 

Thx good to know! 😄 I didn't realise we used the coreos library.

In this case this package is already vendored because helm uses it. I think we should use the same. Medium term the tiller stuff should go away with Helm 3 or a `kubectl apply` based approach.

helmclient.go
rossf7: Ack. I'll add the bool param. I originally planned to use the function when the tunnel to the tiller pod is created. But during testing I found that didn't work. I agree its confusing as it is.

helmclient.go
rossf7: Good call. I'll take a look at splitting it.

helmclient.go
rossf7: I was over thinking the logging I'll remove. 

This return is from a closure that is called by the backoff. So we need to return an error and wait until we are sure tiller is ready. I think this matches the godoc.

helmclient.go
rossf7: I've split out the new code to `installTiller` and `upgradeTiller` methods. More of the existing code could be split out but I think its better to do as separate PRs. I'll continue if I have time.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -86,3 +116,18 @@ func getTillerImage(ctx context.Context, namespace, labelSelector string) (strin
...
 }
+
+func updateTillerImage(ctx context.Context, namespace, labelSelector, tillerImage string) error {
+	d, err := getTillerDeployment(ctx, namespace, labelSelector)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	d.Spec.Template.Spec.Containers[0].Image = tillerImage
+	_, err = config.K8sClient.Apps().Deployments(namespace).Update(d)
```
kopiczko: This feels like a hack. I think we can improve the test a lot with very little effort if we make the `tillerImage` a property of the `Client` struct. This should default to the current value so the change isn't breaking. Then you can install "outdated" tiller with `config.HelmClient.EnsureTillerInstalled` set `config.HelmClient.tillerImage` to the newer version and call `config.HelmClient.EnsureTillerInstalled` again. Such test would better reflect the reality because we don't know what update does under the hood.

rossf7: Yes that is cleaner and simplifies the test. Nice that it means the image is set via EnsureTillerInstalled. I'll update.

helmclient.go
```diff
@@ -57,6 +57,7 @@ type Config struct {
 
 	RestConfig      *rest.Config
 	TillerNamespace string
+	TillerImage     string
```
kopiczko: Giving it a second thought we can even make it an internal. So it's only for changing during the test.

```suggestion
```

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -18,19 +19,55 @@ func TestUpgradeTiller(t *testing.T) {
...
+	tillerImage, err := getTillerImage(ctx, tillerNamespace, labelSelector)
+	if err != nil {
+		t.Fatalf("could not get tiller image %#v", err)
+	}
+	if tillerImage != outdatedTillerImage {
+		t.Fatalf("tiller has not been downgraded got image %#q expected %#q", tillerImage, outdatedTillerImage)
+	}
 
-	_, err = getTillerImage(ctx, namespace, labelSelector)
+	// Upgrade tiller to the latest image using default helm client.
```
kopiczko: I think `config.HelmClient.tillerImage = hlemclient.TillerImageSpec` line is missing here.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -18,19 +19,55 @@ func TestUpgradeTiller(t *testing.T) {
...
+	tillerNamespace := "giantswarm"
+
+	var helmClient *helmclient.Client
+	{
+		c := helmclient.Config{
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
+
+			RestConfig:      config.RestConfig,
+			TillerImage:     outdatedTillerImage,
```
kopiczko: I personally would prefer to create the `helmClient` with the default and change the value just before installing outdated tiller and then revert it. I think the intention would be clearer for the test reader that way. It's easy to miss this detail here IMO.

rossf7:
Upgrade e2e test was failing because there was no check for whether the tiller pod was outdated in EnsureTillerInstalled. Should be OK now.

integration/test/upgradetiller/upgrade_tiller_test.go
rossf7: @kopiczko I looked at this but the tiller image is internal. So I don't see how it can be changed on the fly. I've added a comment about using the outdated image first to make that clearer.

kopiczko: You are right. I made some proposal: https://github.com/giantswarm/helmclient/pull/73

kopiczko: Feel free to merge it to your branch if you find it ok. We can also work on #73 before merging.

helmclient.go
kopiczko: With bool we have to create an error in `newTunnel`. We can also add other cases as needed. Maybe we can handle situation when installed tiller version is newer than expected. `bool` doesn't really work in `newTunnel`.

helmclient.go
```diff
@@ -643,6 +703,12 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 		return nil, microerror.Mask(err)
 	}
 
+	// Do not create a tunnel if tiller is outdated. It will be upgraded.
```
kopiczko:
```suggestion
	// Do not create a tunnel if tiller is outdated.
```

You never know. There is no guarantee that `EnsureTillerInstalled` will be called.

In general I like the idea of not creating tunnel for incompatible tiller version. That solves the problem of talking with updated tiller transparently.

helmclient.go
rossf7: Agreed and its nice that we have the check in `newTunnel`. As we always need a tunnel we don't have to check the tiller version in each method.

helmclient.go
rossf7: Good point. Done.

rossf7:
Yes the install keeps failing with the outdated tiller. The problem with 2.8.1 and 2.8.2 is they have the same tiller version. Only helm client changed and the installer errors rather than upgrades.

I've got a WIP PR to try and get that step of the e2e test to succeed. I'll continue after lunch.
https://github.com/giantswarm/helmclient/pull/74/files

rossf7:
Part of the problem was a typo with the outdated tiller image :( 

There still seems to be a problem with waiting for the pod after tiller is installed. Checking on it.

pipo02mix:
LGTM

rossf7:
Tested with cluster-operator on ginger.

Going with this. Thanks for all the help! ❤️

PR #73, opened 2018-12-12
upgrade-tiller changes

kopiczko:
Note: this is change proposal to #71.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -56,8 +56,8 @@ type Config struct {
 	Logger     micrologger.Logger
 
 	RestConfig      *rest.Config
-	TillerNamespace string
 	TillerImage     string
+	TillerNamespace string
 }
 
 // Client knows how to talk with a Helm Tiller server.
@@ -85,10 +85,10 @@ func New(config Config) (*Client, error) {
 	}
 
 	if config.TillerImage == "" {
-		config.TillerImage = TillerImageSpec
+		config.TillerImage = defaultTillerImage
 	}
 	if config.TillerNamespace == "" {
-		config.TillerNamespace = tillerDefaultNamespace
+		config.TillerNamespace = defaultTillerNamespace
 	}
 
 	c := &Client{
@@ -97,8 +97,9 @@ func New(config Config) (*Client, error) {
 		logger:     config.Logger,
 
 		restConfig:      config.RestConfig,
-		tillerImage:     config.TillerImage,
 		tillerNamespace: config.TillerNamespace,
+
+		tillerImage: defaultTillerImage,
 	}
 
 	return c, nil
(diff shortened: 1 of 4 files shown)


helmclient.go
```diff
@@ -815,23 +816,31 @@ func getTillerImage(pod *corev1.Pod) (string, error) {
...
-		return false, microerror.Mask(err)
+		return microerror.Mask(err)
 	}
 
-	currentTillerVersion, err := parseTillerVersion(currentTillerImage)
+	currentVersion, err := parseTillerVersion(currentImage)
 	if err != nil {
-		return false, microerror.Mask(err)
+		return microerror.Mask(err)
 	}
```
kopiczko: I noticed that below we were using `TillerImageSpec` const which we shouldn't as we specify `tillerImage` field now. This probably breaks some tests.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -19,55 +19,64 @@ func TestUpgradeTiller(t *testing.T) {
 
 	var err error
 
-	labelSelector := "app=helm,name=tiller"
+	currentTillerImage := "quay.io/giantswarm/tiller:v2.8.2"
 	outdatedTillerImage := "gcr.io/kubernetes-helm:v2.7.2"
+
+	labelSelector := "app=helm,name=tiller"
 	tillerNamespace := "giantswarm"
```
kopiczko: Below `currentHelmClient` and `outdatedHelmClient` are set and used in the test.

helmclient.go
kopiczko: Unit tests fixed.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -19,55 +19,64 @@ func TestUpgradeTiller(t *testing.T) {
...
 
-	var helmClient *helmclient.Client
+	var currentHelmClient *helmclient.Client
+	{
+		currentHelmClient = config.HelmClient
+	}
+
+	var outdatedHelmClient *helmclient.Client
```
rossf7: This is nice and yes I agree makes the test easier to read.

helmclient.go
```diff
@@ -815,23 +816,31 @@ func getTillerImage(pod *corev1.Pod) (string, error) {
 	return tillerImage, nil
 }
 
-func isTillerOutdated(pod *corev1.Pod) (bool, error) {
-	currentTillerImage, err := getTillerImage(pod)
+func validateTillerVersion(pod *corev1.Pod, desiredImage string) error {
```
rossf7: Yes changing the method signature back makes sense. As it gets called from newTunnel.

spec.go
```diff
@@ -7,14 +7,12 @@ import (
...
-	// TillerImageSpec is the tiller image to be installed.
-	TillerImageSpec = "quay.io/giantswarm/tiller:v2.8.2"
-
 	// defaultMaxHistory is the maximum number of release versions stored per
 	// release by default.
 	defaultMaxHistory = 10
 
-	tillerDefaultNamespace = "kube-system"
+	defaultTillerImage     = "quay.io/giantswarm/tiller:v2.8.2"
+	defaultTillerNamespace = "kube-system"
```
rossf7: <3

rossf7:
> Feel free to merge it to your branch if you find it ok. We can also work on #73 before merging.

Sorry I missed this. Going to merge and will continue with testing.

PR #77, opened 2019-01-04
Add helmclienttest package

rossf7:

This is prep for some helmclient changes I need to make for chart-operator. Adds the helmclienttest package so we can remove the `mock_test.go` files we have in chart-operator and cluster-operator.

e.g. https://github.com/giantswarm/chart-operator/blob/1553942b22cbecea7b26b7b90bb56399d3b3d8f6/service/controller/chartconfig/v5/resource/chart/mock_test.go

### DIFF
--- a/helmclienttest/helmclient.go
+++ b/helmclienttest/helmclient.go
@@ -1,0 +1,78 @@
+package helmclienttest
+
+import (
+	"context"
+
+	"github.com/giantswarm/helmclient"
+	"k8s.io/helm/pkg/helm"
+)
+
+type Config struct {
+	DefaultReleaseContent *helmclient.ReleaseContent
+	DefaultReleaseHistory *helmclient.ReleaseHistory
+	DefaultError          error
+}
+
+type Client struct {
+	defaultReleaseContent *helmclient.ReleaseContent
+	defaultReleaseHistory *helmclient.ReleaseHistory
+	defaultError          error
+}
+
+func New(config Config) (helmclient.Interface, error) {
+	c := &Client{
+		defaultReleaseContent: config.DefaultReleaseContent,
+		defaultReleaseHistory: config.DefaultReleaseHistory,
+		defaultError:          config.DefaultError,
+	}
+
+	return c, nil
+}
+
+func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options ...helm.DeleteOption) error {
+	if c.defaultError != nil {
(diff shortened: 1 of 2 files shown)


helmclienttest/helmclient.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+
+	"github.com/giantswarm/helmclient"
+	"k8s.io/helm/pkg/helm"
+)
+
+type Config struct {
+	DefaultReleaseContent *helmclient.ReleaseContent
+	DefaultReleaseHistory *helmclient.ReleaseHistory
+	DefaultError          error
```
kopiczko: I like having a mock in general. But how about changing the approach and make it more flexible? Example below:

```go
type Config struct {
	GetReleaseContentFunc func(ctx context.Context, releaseName string) (*helmclient.ReleaseContent, error)
}
```

And then:

```go
func (c *Client) GetReleaseContent(ctx context.Context, releaseName string) (*helmclient.ReleaseContent, error) {
	if c.getReleaseContentFunc != nil {
		return c.GetReleaseContentFunc()
	}

	return nil, nil
}
```

Such approach would allow to implement anything within the mock w/o having to implement all the methods.

rossf7: Yes that is nice. I don't want to spend lots of time on this but the approach is more flexible. I'll look at it later.

kopiczko: Yes, but then we'll have to spend more time on vendoring. You can expose only functions that will be used by you in testing. Extending the object is backward compatible.

rossf7: @kopiczko Looking at this again. Its more flexible but it creates another problem.

https://github.com/giantswarm/chart-operator/blob/1553942b22cbecea7b26b7b90bb56399d3b3d8f6/service/controller/chartconfig/v5/resource/chart/mock_test.go#L50-L56
https://github.com/giantswarm/cluster-operator/blob/04061930d9404bd9b406d68437de3b1c46d31c6f/pkg/v10/chartconfig/mock_test.go#L46-L52

The mocks are the same in multiple projects but by passing in `GetReleaseContentFunc` we need to define the same functions in both. I think its better to extract all the mock logic to this package. WDYT?

kopiczko: Ack. I didn't realize this is already in use. Let's leave it as it is.

PR #79, opened 2019-01-09
Add LoadChart to client

rossf7:

Adds LoadChart to get chart metadata which wraps the upstream [Load](https://github.com/helm/helm/blob/4634f5f2faa0caa7de390c95c7d27be22d56a156/pkg/chartutil/load.go#L38-L58) method.

> Load takes a string name, tries to resolve it to a file or directory, and then loads it.
> 
> This is the preferred way to load a chart. It will discover the chart encoding
> and hand off to the appropriate chart reader.

It returns the protobufs struct. So I've added a `key` package for getting values. I need this in chart-operator to get the chart version. We'll also need it in the front end for getting things like icon URLs.

This is to break down #78 which also adds `PullChartTarball` and has got a bit big. An e2e test will be added in that PR.


key/key.go
```diff
@@ -0,0 +1,13 @@
...
+	hapichart "k8s.io/helm/pkg/proto/hapi/chart"
+)
+
+func ChartVersion(chart *hapichart.Chart) string {
+	if chart.Metadata == nil {
+		return ""
+	}
+
+	return chart.Metadata.Version
+}
```
rossf7: Just adding ChartVersion for now. We can add more helpers as needed.

kopiczko: I would prefer to return custom tailored type instead of having extractors here. Something like https://github.com/giantswarm/devctl/blob/be38aacbff4b0409211ad2fd36e5803b0f99cc58/pkg/githubclient/types.go#L11.

We could have a `Chart` type with `Version` field. The usage would be obvious then. We can add fields as we need then and don't need to import anything outside library.

rossf7: Yes that's how I had it originally. The problem is the `Metadata` struct is large and we will need a lot of values from it. This is because we want the `Chart.yaml` to be the master copy for metadata about charts.

https://github.com/helm/helm/blob/4634f5f2faa0caa7de390c95c7d27be22d56a156/pkg/proto/hapi/chart/metadata.pb.go#L75-L112

rossf7: I'd like to get more feedback on this. @kopiczko, your point on not needing to import any helm packages is a good one. My concern is just that we'll need to end up replicating the protobufs struct.

kopiczko: When I was developing this github package I also did a lot of nil checking (and possibly other validation) which is madness with protobuf generated code so we don't have to repeat everywhere. Something you have there `chart == nil || chart.Metadata == nil`.

Will we ever use more than 10% of those fields? And is it really an overhead with maintaining a struct when we have an extractor for everything anyway? I think it is actually easier. It's super easy to get lost in all that extractors. We every so often notice that we don't use some for a long time in our operators.

rossf7: We will definitely use more than 10% of the metadata fields. But I agree the protobuf structs are hard to work with. Cool I'll define a struct in spec.go and go with that approach.

kopiczko: I'd prefer (kind of obvious - this is how I did it) to have that in `types.go`. I don't see why would we pollute `spec.go` with private constructors and validations. `spec.go` declare returned types and I think it's enough to follow what's going on. Especially given that we often return types from outside the package.

rossf7:
@kopiczko PTAL.

types.go
```diff
@@ -0,0 +1,25 @@
+package helmclient
+
+// Chart returns information about a Helm Chart.
+type Chart struct {
+	// Version is the version of the Helm Chart.
+	Version string
+}
+
```
kopiczko: Are the structs below used?

helmclient.go
```diff
@@ -541,6 +541,24 @@ func (c *Client) ListReleaseContents(ctx context.Context) ([]*ReleaseContent, er
 	return contents, nil
 }
 
+// LoadChart loads a Helm Chart and returns relevant parts of its structure.
+func (c *Client) LoadChart(ctx context.Context, chartPath string) (*Chart, error) {
```
kopiczko: I think it should return a value. This is just a data type.

```suggestion
func (c *Client) LoadChart(ctx context.Context, chartPath string) (Chart, error) {
```

helmclient.go
```diff
@@ -541,6 +541,24 @@ func (c *Client) ListReleaseContents(ctx context.Context) ([]*ReleaseContent, er
...
+	helmChart, err := chartutil.Load(chartPath)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
```
kopiczko: Nit: I'd put the stuff below in the `newChart` constructor.

helmclienttest/helmclient.go
```diff
@@ -65,6 +69,14 @@ func (c *Client) ListReleaseContents(ctx context.Context) ([]*helmclient.Release
 	return nil, nil
 }
 
+func (c *Client) LoadChart(ctx context.Context, chartPath string) (*helmclient.Chart, error) {
```
kopiczko:
```suggestion
func (c *Client) LoadChart(ctx context.Context, chartPath string) (Chart, error) {
```

helmclienttest/helmclient.go
```diff
@@ -3,27 +3,31 @@ package helmclienttest
...
 
-	"github.com/giantswarm/helmclient"
 	"k8s.io/helm/pkg/helm"
+
+	"github.com/giantswarm/helmclient"
 )
 
 type Config struct {
+	DefaultChart          *helmclient.Chart
+	DefaultError          error
```
kopiczko: How about making that:

```
	LoadChartResponse Chart
	LoadChartError    error
```

and deprecating fields below?

Not as flexible as having `LoadChartFunc` but simpler to setup and fits better what we have now.

helmclient.go
rossf7: OK makes sense. Done.

helmclienttest/helmclient.go
rossf7: Done.

types.go
rossf7: Yes they are used by GetReleaseContent and GetReleaseHistory.

helmclient.go
rossf7: I don't follow here. You mean add a newChart function?

kopiczko: Something like this https://github.com/giantswarm/devctl/blob/be38aacbff4b0409211ad2fd36e5803b0f99cc58/pkg/githubclient/types.go#L18

helmclienttest/helmclient.go
kopiczko: Cool I think `DefaultChart` can be removed now.

types.go
kopiczko: Ah sure they are moved from `spec.go`. I haven't noticed that in the first place 👍

helmclienttest/helmclient.go
rossf7: Yup missed that. Gone now thx!

PR #80, opened 2019-01-10
Add PullChartTarball

rossf7:

So my helmclient PRs continue! This is the last method I need to add and is based on what we do in `apprclient`. I need to add this to unblock my chart-operator changes.

I'm working on an e2e test in #78 but its got a bit big so I'll do a separate PR.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"strings"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/giantswarm/k8sportforward"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -25,11 +27,6 @@ import (
 	hapichart "k8s.io/helm/pkg/proto/hapi/chart"
 	hapirelease "k8s.io/helm/pkg/proto/hapi/release"
 	hapiservices "k8s.io/helm/pkg/proto/hapi/services"
-)
-
-const (
-	// runReleaseTestTimeout is the timeout in seconds when running tests.
-	runReleaseTestTimout = 300
 )
 
 var (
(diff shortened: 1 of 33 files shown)


helmclient.go
```diff
@@ -27,11 +29,6 @@ import (
 	hapiservices "k8s.io/helm/pkg/proto/hapi/services"
 )
 
-const (
-	// runReleaseTestTimeout is the timeout in seconds when running tests.
-	runReleaseTestTimout = 300
-)
-
```
rossf7: I moved this to spec.go. I think its better to have all the constants together.

spec.go
```diff
@@ -10,6 +10,10 @@ const (
 	// defaultMaxHistory is the maximum number of release versions stored per
 	// release by default.
 	defaultMaxHistory = 10
+	// httpClientTimeout is the timeout when pulling tarballs.
+	httpClientTimeout = 5
```
rossf7: This constant is new.

pull_chart_tarball.go
```diff
@@ -0,0 +1,90 @@
+package helmclient
```
rossf7: Using a separate file as helmclient.go is nearly 1k LOC.

pull_chart_tarball.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			buf := new(bytes.Buffer)
+			_, err = buf.ReadFrom(resp.Body)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+			return microerror.Maskf(invalidStatusCodeError, fmt.Sprintf("got StatusCode %d with body %s", resp.StatusCode, buf.String()))
```
kopiczko: I don't think we handle that anywhere.

```suggestion
			return microerror.Maskf(executionFailedError, fmt.Sprintf("got StatusCode %d with body %s", resp.StatusCode, buf.String()))
```

pull_chart_tarball.go
kopiczko: I personally don't mind having it even 3k loc.

pull_chart_tarball.go
rossf7: Good point. Done.

pull_chart_tarball.go
rossf7: I prefer the separate file since it only relates to PullChartTarball. I think doing the same for EnsureTillerInstalled would be nice as its also a lot of LOC.

PR #83, opened 2019-01-14
Improve tiller version check when deployment has been changed manually

rossf7:
We had a problem in `talos` because the Tiller deployment had been changed manually and didn't have a version tag `quay.io/giantswarm/tiller`. 

```
panic: [{/go/src/github.com/giantswarm/draughtsman/service/service.go:171: service.Boot} {/go/src/github.com/giantswarm/draughtsman/vendor/github.com/giantswarm/helmclient/helmclient.go:278: } {/go/src/github.com/giantswarm/draughtsman/vendor/github.com/giantswarm/helmclient/helmclient.go:863: } {/go/src/github.com/giantswarm/draughtsman/vendor/github.com/giantswarm/helmclient/helmclient.go:886: tiller image `quay.io/giantswarm/tiller` is invalid} {execution failed error}]
```

Currently we error and this causes a panic and alerts. This is to make the version check smarter. If the image tag is empty or `latest` we treat it as outdated. So the tiller deployment gets upgraded to have the correct version.

See https://gigantic.slack.com/archives/C03CPNRTS/p1547479149027500.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -880,13 +880,26 @@ func validateTillerVersion(pod *corev1.Pod, desiredImage string) error {
 }
 
 func parseTillerVersion(tillerImage string) (*semver.Version, error) {
+	defaultVersion, err := semver.NewVersion("0.0.0")
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
 	// Tiller image tag has the version.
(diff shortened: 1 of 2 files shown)


rossf7:
Yes its not very pretty but should prevent alerts when the deployment gets changed manually.

PR #85, opened 2019-01-16
Add AppVersion and LastDeployed to GetReleaseHistory

rossf7:

See https://github.com/giantswarm/chart-operator/pull/183#issue-245609254 for more details.


helmclient.go
```diff
@@ -430,14 +432,26 @@ func (c *Client) GetReleaseHistory(ctx context.Context, releaseName string) (*Re
 	{
 		release := resp.Releases[0]
 
+		var appVersion string
 		var version string
 		if release.Chart != nil && release.Chart.Metadata != nil {
+			appVersion = release.Chart.Metadata.AppVersion
```
rossf7: We also get the AppVersion for app-operator metrics.

helmclient.go
```diff
@@ -430,14 +432,26 @@ func (c *Client) GetReleaseHistory(ctx context.Context, releaseName string) (*Re
...
 			version = release.Chart.Metadata.Version
 		}
 
+		var lastDeployed time.Time
+		if release.Info != nil {
+			lastDeployed, err = ptypes.Timestamp(release.Info.LastDeployed)
```
rossf7: This is the protocol buffers magic to convert the timestamp field to a Go time.Time.

helmclient.go
```diff
@@ -430,14 +432,26 @@ func (c *Client) GetReleaseHistory(ctx context.Context, releaseName string) (*Re
...
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+		}
+
 		history = &ReleaseHistory{
```
xh3b4sd: FMI what is the term history about? Name and version are not so much historical, isn't it?

types.go
```diff
@@ -18,6 +20,10 @@ type ReleaseContent struct {
 
 // ReleaseHistory returns version information about a Helm Release.
 type ReleaseHistory struct {
+	// AppVersion is the app version of the Helm Chart that has been deployed.
```
xh3b4sd: Can we have examples for the versions? Or are all of these only semver? I cannot really make sense of this without understanding the whole stack. For instance, I have issues understanding what the difference between the app version and version of helm chart actually is. People told me already but my brain forgot. :D

kopiczko: I also don't understand the difference between that and the `Version`.

puja108: with our own components we can sync our version of the component and the chart version, however, with 3rd party apps like prometheus for example there's a version of the app like 2.6.0 (for prom) and a version of our chart that changes separately from the appVersion, e.g. if we change some flag or other stuff around the app. current example would be that our NGINX-IC chart version is at 0.3.2 and the appVersion is at 0.21.0.

We might be able to phrase that in a clearer way in the comments and maybe also mention that for our own components this is the same.

btw, @rossf7 as our charts often include more than one component this appVersion can only be the main component like prom here (grafana and exporters and stuff are not shown here). I think that's ok, but maybe that's also sth to clarify, that appVersion is the version of the main app in the app package

rossf7: See https://github.com/giantswarm/apiextensions/pull/190#discussion_r249089157 I also updated the godoc here. WDYT?

helmclient.go
rossf7: The naming is aligned with the upstream helm method that we call.

`c.newHelmClientFromTunnel(t).ReleaseHistory(releaseName, helmclient.WithMaxHistory(1))`

types.go
kopiczko: I think this property sucks because it allows to specify only a single component :D 
And it only makes sense when there are multiple...

But that's just a rant.

rossf7: Yes it's maybe a bit messy but it can be useful and its what's available via the Helm Chart format.

PR #88, opened 2019-02-18
Wait for tiller to be installed

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/5218

Fixes to helmclient needed for https://github.com/giantswarm/cluster-operator/pull/386. With this all chartconfig CRs and chart-operator are installed in a single reconciliation loop the first time the tenant cluster becomes available.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -271,7 +271,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		}
 	}
 
-	if pod != nil {
+	if !installTiller && pod != nil {
 		err = validateTillerVersion(pod, c.tillerImage)
 		if IsTillerOutdated(err) {
 			upgradeTiller = true
--- a/helmclient.go
+++ b/helmclient.go
@@ -262,7 +262,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 
 			return nil
 		}
-		b := backoff.NewExponential(2*time.Minute, 5*time.Second)
+		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
 		n := backoff.NewNotifier(c.logger, context.Background())
 
 		err := backoff.RetryNotify(o, b, n)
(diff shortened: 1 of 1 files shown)


helmclient.go
```diff
@@ -271,7 +271,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		}
 	}
 
-	if pod != nil {
+	if !installTiller && pod != nil {
```
rossf7: This fixes a bug I introduced with the tiller upgrade logic. If we are installing tiller there is no point doing the version check.

This was causing this error to be returned in some situations.

`return microerror.Maskf(executionFailedError, "invalid state cannot both install and upgrade tiller")`

helmclient.go
```diff
@@ -313,7 +313,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 
 		o := func() error {
 			t, err := c.newTunnel()
-			if IsTillerNotFound(err) {
+			if !installTiller && IsTillerNotFound(err) {
```
rossf7: This is the fix needed for https://github.com/giantswarm/cluster-operator/pull/386. Currently we always do a permanent backoff when installing tiller. This slows the bootstrap process as we need to wait for the resync period.

helmclient.go
```diff
@@ -262,7 +262,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 
 			return nil
 		}
-		b := backoff.NewExponential(2*time.Minute, 5*time.Second)
+		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
```
rossf7: Current backoff is too long. I want to reduce it but not so far that it causes problems with e2e.

helmclient.go
```diff
@@ -333,7 +333,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 
 			return nil
 		}
-		b := backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval)
+		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
```
rossf7: Again reducing the backoff.

xh3b4sd:
Random brain fart. We think about extending the resource interfaces with the upstream [reconcile.Result](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/reconcile#Result). For these cases we could then say we want to re-queue after 10 seconds and not wait for the full 5 minutes. This primitive is pretty neat. Let's keep it in mind and get going with it after VOO. /cc @kopiczko @tuommaki

PR #90, opened 2019-03-08
Fix check when both install and upgrade tiller are true

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/5224

I've seen this error `invalid state cannot both install and upgrade tiller` a few times during cluster creation when cluster-operator is bootstrapping tiller.

This fixes the current check which is just wrong. 🤦‍♂️

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -299,7 +299,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		if err != nil {
 			return microerror.Mask(err)
 		}
-	} else {
+	} else if installTiller && upgradeTiller {
 		return microerror.Maskf(executionFailedError, "invalid state cannot both install and upgrade tiller")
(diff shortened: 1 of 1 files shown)


helmclient.go
```diff
@@ -299,7 +299,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		if err != nil {
 			return microerror.Mask(err)
 		}
-	} else {
+	} else if installTiller && upgradeTiller {
```
rossf7: Currently this error is returned if both installTiller and upgradeTiller are false. But this is a valid case. 

installTiller is false because the deployment has already been created but the pod is not ready yet.
We should continue and wait for the ready pod which is the next step. 

Now we just error if both flags are true. This should not be possible but I think its good to error as it means something is very wrong.

PR #91, opened 2019-03-08
Add value input to tiller installer

MarcelMue:
Required to augment the tiller deployment and add priorityclasses, tolerations etc pp.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -262,7 +262,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 
 			return nil
 		}
-		b := backoff.NewExponential(2*time.Minute, 5*time.Second)
+		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
 		n := backoff.NewNotifier(c.logger, context.Background())
 
 		err := backoff.RetryNotify(o, b, n)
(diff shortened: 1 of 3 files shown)


helmclient.go
```diff
@@ -286,6 +295,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 		MaxHistory:                   defaultMaxHistory,
 		Namespace:                    c.tillerNamespace,
 		ServiceAccount:               tillerPodName,
+		Values:                       values,
```
MarcelMue: This is all that changes, very straight forward just passing this through to the helm API.

helmclient.go
```diff
@@ -153,6 +153,15 @@ func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options
 // As a first step, it checks if Tiller is already ready, in which case it
 // returns early.
 func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
+	return c.EnsureTillerInstalledWithValues(ctx, []string{})
```
MarcelMue: Works fine and preserves the behaviour.

MarcelMue:
@rossf7 `chart-operator` for now, so we get the correct priority class and so on in tenant clusters.

rossf7:
Ack. But tiller in tenant clusters is first created by the tiller resource in cluster-operator. This is for bootstrapping because we install chart-operator with helm.

https://github.com/giantswarm/cluster-operator/blob/4be49f490f2548ffe82c6ec0f1c730265cf9eb86/pkg/v13/resource/tiller/resource.go#L68

MarcelMue:
@rossf7 Argh I see - I was starting to wonder why it wasn't working from the chart-operator already ...
I will add my changes to cluster-operator

rossf7:
Cool I think its good to have it in both.

PR #92, opened 2019-03-18
reduce backoff to prevent blocking reconciliation

xh3b4sd:
This is probably a more controversial change. Do you think this makes sense? Will this break something or cause unintended changes we can anticipate already? I saw cases in which tenant cluster creation got delayed dramatically because other clusters failing blocked reconciliation of the new clusters for longer periods of time.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -137,7 +137,7 @@ func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options 
 
 		return nil
 	}
-	b := backoff.NewMaxRetries(3, 1*time.Second)
+	b := backoff.NewMaxRetries(5, 1*time.Second)
 	n := backoff.NewNotifier(c.logger, ctx)
 
 	err := backoff.RetryNotify(o, b, n)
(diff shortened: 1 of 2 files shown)


kopiczko:
I thought about this in general. I thought we could add `wait bool` arg to all functions using backoff. Then we can use it both as call-once-and-error-on-failure or call-till-it-works  for reconciliation.

As for the consequences I'd check e2e tests.

xh3b4sd:
The e2e tests are sorted. I do not want to increase the complexity of the helm client magic. It already works around clear primitives. We should rather split it up further and implement decision making strategies where they are defined by the business domain. The e2e test handling should do what the tests are concerned about. The resource implementations should do what the operators are concerned about. There should only be a simple primitive like "create" or "update" this thing. When the actions of created or updated state in the system cannot be completed synchronously, we should accept that fact. Then requesting the work to be done and coming back later is the way to go IMO.

xh3b4sd:
I would like I would like I would like to get more feeheeeedaback.

xh3b4sd:
For the time being it should then be the resource implementation that calls `EnsureTillerInstalled` until it is happy. Can you point to the code?

rossf7:
@xh3b4sd Cool yes the resource taking care makes sense to me. Code is here.

https://github.com/giantswarm/cluster-operator/blob/4be49f490f2548ffe82c6ec0f1c730265cf9eb86/pkg/v13/resource/tiller/resource.go#L86

kopiczko:
How about having `RetryClient` wrapping original client and dropping backoff completely?

xh3b4sd:
That would work when you could do it per primitive but you would probably configure one to rule them all. The provided functionalities are too complex and you probably want to handle it per situation, just like the `cluster-operator` requires the special handling only for ensuring Tiller is installed. I would provide PRs for `cluster-operator` and aim for getting the backoffs down here.

PR #93, opened 2019-04-02
Add priorityclass to ensure it exists when installing tiller

MarcelMue:
Follow up to https://github.com/giantswarm/chart-operator/pull/206

Priority class is the same as https://github.com/giantswarm/k8scloudconfig/blob/100d834280834beb78233661a6a6d7aeecba7749/v_4_1_1/files/k8s-resource/priority_classes.yaml

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -258,6 +258,8 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 	{
 		priorityClassName := "giantswarm-critical"
 
+		c.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating priorityclass %#q", priorityClassName))
+
 		p := &schedulingv1beta1.PriorityClass{
 			TypeMeta: metav1.TypeMeta{
 				APIVersion: "scheduling.k8s.io/v1alpha1",
@@ -271,7 +273,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			Description:   "This priority class is used by giantswarm kubernetes components.",
 		}
 
-		_, err := c.k8sClient.SchedulingV1alpha1().PriorityClasses().Create(p)
+		_, err := c.k8sClient.SchedulingV1beta1().PriorityClasses().Create(p)
 		if errors.IsAlreadyExists(err) {
 			c.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("priorityclass %#q already exists", priorityClassName))
 			// fall through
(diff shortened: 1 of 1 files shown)


helmclient.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	schedulingv1alpha1 "k8s.io/api/scheduling/v1alpha1"
```
rossf7: This errored but switching to v1beta1 fixed it. This was with k8s 1.13 and kind.

helmclient.go
```diff
@@ -253,6 +254,35 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 		}
 	}
 
+	// Create the priority class for tiller to make sure it is always deployed.
+
```
rossf7: Nitty but no need for a newline here IMO.

helmclient.go
MarcelMue: Done

helmclient.go
MarcelMue: Removed.

MarcelMue:
I think e2e is failing because we don't have a recent enough k8s version to use `v1beta1`?

Unsure right now but it complains with `{/go/src/github.com/giantswarm/helmclient/helmclient.go:281: } {the server could not find the requested resource}` when we try to create the priority class.

MarcelMue:
Yeah I think the minikube is 1.10 and this is newer :(

rossf7:
Damn yes I agree I think its minikube :(

MarcelMue:
I tried to quickly get this in by using minikube 1.0.0
It crashes immediately unfortunately :( I don't really want to invest too much time here and instead will work more towards kind

https://github.com/giantswarm/e2e-harness/pull/191

rossf7:
@MarcelMue Thanks for trying and no worries. I agree better to spend the energy on moving to kind than fighting with minikube.

sslavic:
@MarcelMue @rossf7 noticed this PR while working now on k8s-1-16 update.

We have migrated helmclient e2e tests to use kind/e2ectl - so likely this work can be continued and shipped, if it's still relevant.

rossf7:
@sslavic Thanks for the heads up. This is mainly useful when running operators locally. As in control plane and tenant clusters the priority class is created via ignition.

@MarcelMue OK if I assign myself and I'll finish this once the k8s 1.16 changes are done?

MarcelMue:
@rossf7 Yes, sure. This got left behind on my plate unfortunately.

sslavic:
In various operator e2e tests I was dealing with (something similar to) this by conditionally (only in e2e tests) adding the priority class resource in operator's helm chart. I think we also may have few occurrences where preconditions like this were created via circle+kubectl. Maybe this PR would render those hacks as no longer needed so they could be cleaned up.

rossf7:
Closing. Done in #145

PR #94, opened 2019-04-04
Add support for merging values

rossf7:

Adds support for merging values. This will be used in chart-operator, app-operator and opsctl for merging values when installing or updating Helm releases.

Logic is based on the upstream merge logic in helm. See other comments for the link.


merge_values.go
```diff
@@ -7,6 +7,93 @@ import (
...
+// byte array. Only a single key is supported and the input data structure
+// matches the configmap or secret where the data is stored.
+func processYAML(inputMap map[string][]byte) (map[string]interface{}, error) {
+	var err error
+
+	result := map[string]interface{}{}
+
+	if len(inputMap) > 1 {
+		return nil, microerror.Maskf(executionFailedError, "merging %d keys is unsupported expected 1 key", len(inputMap))
+	}
```
rossf7: Only a single key with YAML is supported. But the key can have any name so there are no magic strings.

merge_values.go
```diff
@@ -7,6 +7,93 @@ import (
...
+
+	return result, nil
+}
+
+// mergeValues implements the merge logic. It performs a deep merge. If a value
+// is present in both then the source map is preferred.
+//
+// Logic is based on the upstream logic implemented by Helm.
+// https://github.com/helm/helm/blob/240e539cec44e2b746b3541529d41f4ba01e77df/cmd/helm/install.go#L358
```
rossf7: See this link for the upstream code which the merge logic is based on. I just made minor changes to follow our usual conventions.

merge_values_test.go
```diff
@@ -41,13 +41,134 @@ nested:
   deeper:
     bottom: true
     float: 274877.906944
+test: test`
+	overrideNestedYaml = `
```
xh3b4sd: What about having separate constant blocks for these things? Took me a while to figure what is going on here.

rossf7: Good point that is easier to read. Thanks, done.

PR #95, opened 2019-04-11
Reduce backoff exponential time & catch YAML parsing error as permanent

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/5743

Only affect `InstallReleaseFromTarball` function.

### DIFF
--- a/error.go
+++ b/error.go
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
 func IsTooManyResults(err error) bool {
 	return microerror.Cause(err) == tooManyResultsError
 }
+
+var (
+	yamlConvertingFailedRegexp = regexp.MustCompile(`error converting YAML to JSON:`)
+)
+
+var yamlConvertingFailedError = &microerror.Error{
+	Kind: "releaseAlreadyExistsError",
+}
+
+// IsYamlConvertingFailed asserts yamlConvertingFailedError.
+func IsYamlConvertingFailed(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	c := microerror.Cause(err)
+
+	if c == yamlConvertingFailedError {
+		return true
+	}
+	if yamlConvertingFailedRegexp.MatchString(c.Error()) {
+		return true
+	}
+
+	return false
+}
--- a/error.go
+++ b/error.go
@@ -223,7 +223,7 @@ func IsTooManyResults(err error) bool {
 }
 
 var (
-	yamlConvertingFailedPrefix = "error converting YAML to JSON:"
+	yamlConversionFailedErrorText = "error converting YAML to JSON:"
 )
 
 var yamlConvertingFailedError = &microerror.Error{
@@ -241,7 +241,7 @@ func IsYamlConvertingFailed(err error) bool {
 	if c == yamlConvertingFailedError {
 		return true
 	}
-	if strings.Contains(c.Error(), yamlConvertingFailedPrefix) {
+	if strings.Contains(c.Error(), yamlConversionFailedErrorText) {
 		return true
 	}
 
--- a/error.go
+++ b/error.go
@@ -223,11 +223,11 @@ func IsTooManyResults(err error) bool {
 }
 
 var (
-	yamlConvertingFailedRegexp = regexp.MustCompile(`error converting YAML to JSON:`)
+	yamlConvertingFailedPrefix = "error converting YAML to JSON:"
 )
 
 var yamlConvertingFailedError = &microerror.Error{
-	Kind: "releaseAlreadyExistsError",
+	Kind: "yamlConvertingFailedError",
 }
 
 // IsYamlConvertingFailed asserts yamlConvertingFailedError.
@@ -241,7 +241,7 @@ func IsYamlConvertingFailed(err error) bool {
 	if c == yamlConvertingFailedError {
 		return true
 	}
-	if yamlConvertingFailedRegexp.MatchString(c.Error()) {
+	if strings.Contains(c.Error(), yamlConvertingFailedPrefix) {
 		return true
 	}
 
--- a/error.go
+++ b/error.go
@@ -226,19 +226,19 @@ var (
 	yamlConversionFailedErrorText = "error converting YAML to JSON:"
 )
 
-var yamlConvertingFailedError = &microerror.Error{
+var yamlConversionFailedError = &microerror.Error{
 	Kind: "yamlConvertingFailedError",
 }
 
 // IsYamlConvertingFailed asserts yamlConvertingFailedError.
-func IsYamlConvertingFailed(err error) bool {
+func IsYamlConversionFailed(err error) bool {
 	if err == nil {
 		return false
 	}
 
 	c := microerror.Cause(err)
 
-	if c == yamlConvertingFailedError {
+	if c == yamlConversionFailedError {
 		return true
 	}
 	if strings.Contains(c.Error(), yamlConversionFailedErrorText) {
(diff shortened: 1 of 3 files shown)


rossf7:
Please check the other CRUD methods for helm releases. AIUI `UpdateReleaseFromTarball` was where the problem was in the PM.

helmclient.go
```diff
@@ -137,7 +137,7 @@ func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options
 
 		return nil
 	}
-	b := backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval)
+	b := backoff.NewExponential(30*time.Second, 5*time.Second)
```
kopiczko: Doesn't it make sense to call max tries instead? There is no really much room for exponential backoffs.

tomahawk28: WDYT as below? ```
b := backoff.NewMaxRetries(5, 10*time.Second)```

kopiczko: I'd rather say `(10, 5*time.Second)`. 5 seconds is the max with your previous proposal. Have you tested that is enough?

tomahawk28: I will try to do that in `ginger` after an agreement.

rossf7: We want the backoff to be short so other CRs can be processed. But not so short that the operator can't do what it needs to do. Pawel's proposal lgtm but it needs to be tested.

I would test this with the Ingress Controller chart as it is complex. So check it can be created, updated and deleted with these settings.

helmclient.go
```diff
@@ -660,6 +664,8 @@ func (c *Client) UpdateReleaseFromTarball(ctx context.Context, releaseName, path
 		release, err := c.newHelmClientFromTunnel(t).UpdateRelease(releaseName, path, options...)
 		if IsReleaseNotFound(err) {
 			return backoff.Permanent(microerror.Mask(err))
+		} else if IsYamlConvertingFailed(err) {
```
tomahawk28: To mitigate the converting issue stop making other CR reconciliation, I add this new error checking function.

tomahawk28:
@rossf7 @kopiczko PTAL, if it's okay I would like to test again.

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
 func IsTooManyResults(err error) bool {
 	return microerror.Cause(err) == tooManyResultsError
 }
+
+var (
+	yamlConvertingFailedRegexp = regexp.MustCompile(`error converting YAML to JSON:`)
```
rossf7:
```suggestion
	yamlConversionFailedRegexp = regexp.MustCompile(`error converting YAML to JSON:`)
```

I think this naming is clearer. Can you move regex to the top of the file like in the errors library.

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
...
+)
+
+var yamlConvertingFailedError = &microerror.Error{
+	Kind: "releaseAlreadyExistsError",
```
rossf7: Name is wrong.

error.go
xh3b4sd: I think there is no reason for a regular expression. It looks like you only need to check for a string match. https://golang.org/pkg/strings/#Contains is your friend.

xh3b4sd:
Can you give some data on how much time we spent before the PR and how much we spent with the PR?

tomahawk28:
@xh3b4sd  Before we apply this change, each CR takes 2 minutes in helmclient related operation, and `operatorkit` also tried it 3 times, so should be **6 minutes** even for fatal tiller error. 
With this, it took 50 seconds for each CR (10 times * 5) and 3 times in operatorkit, so it should reduce to **150 seconds**.

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
 func IsTooManyResults(err error) bool {
 	return microerror.Cause(err) == tooManyResultsError
 }
+
+var (
+	yamlConvertingFailedPrefix = "error converting YAML to JSON:"
```
rossf7:
```suggestion
	yamlConversionFailedErrorText = "error converting YAML to JSON:"
```

As I said in my previous review I prefer `Conversion` over `Converting`.  As it matches our usual naming conventions.

Also you're using `strings.Contains` not `strings.HasPrefix`. So I think having prefix in the name is confusing.

tomahawk28: I doubt that I could use `HasPrefix` since the keywords would be located in the middle of the sentence. 
```rpc error: code = Unknown desc = YAML parse error on kubernetes-nginx-ingress-controller-chart/templates/configmap.yaml: error converting YAML to JSON: yaml: line 16: did not find expected key```
Maybe we should change the name of `prefix` to something else?

Comment (rossf7) [reply error.go]: Exactly this isn't a prefix because its not at the start of the error message. That's why I suggested using `ErrorText` in the name. 

https://dictionary.cambridge.org/grammar/british-grammar/word-formation/prefixes

**Comment by tomahawk28** [comment]
Tested on `ginger`. 

1. Deleted two `chartconfigs` CR and put some error-prone `configmaps` related to nginx. 
2. Created two `chartconfigs` and check whether it passes to other CR immediately after one failure.
3. Checked that it skips nginx chart reconciliation after 3 retries (it's on operatorkit), and skip to net-exporter chart.

File: error.go [hunk 6]
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
 func IsTooManyResults(err error) bool {
 	return microerror.Cause(err) == tooManyResultsError
 }
+
+var (
+	yamlConversionFailedErrorText = "error converting YAML to JSON:"
+)
+
+var yamlConvertingFailedError = &microerror.Error{
```
rossf7:
```suggestion
var yamlConversionFailedError = &microerror.Error{
```

I guess it wasn't clear but you only changed the line I highlighted. The rest of the PR needs to be updated too.

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
...
+
+var (
+	yamlConversionFailedErrorText = "error converting YAML to JSON:"
+)
+
+var yamlConversionFailedError = &microerror.Error{
+	Kind: "yamlConvertingFailedError",
+}
+
+// IsYamlConvertingFailed asserts yamlConvertingFailedError.
```
rossf7:
```suggestion
// IsYamlConversionFailed asserts yamlConversionFailedError.
```

tomahawk28: I'm sorry, have a tough morning.

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
 func IsTooManyResults(err error) bool {
 	return microerror.Cause(err) == tooManyResultsError
 }
+
+var (
+	yamlConversionFailedErrorText = "error converting YAML to JSON:"
+)
+
+var yamlConversionFailedError = &microerror.Error{
+	Kind: "yamlConvertingFailedError",
```
rossf7:
```suggestion
	Kind: "yamlConversionFailedError",
```

error.go
```diff
@@ -221,3 +221,29 @@ var tooManyResultsError = &microerror.Error{
...
+}
+
+// IsYamlConversionFailed asserts yamlConvertingFailedError.
```
rossf7:
```suggestion
// IsYamlConversionFailed asserts yamlConversionFailedError.
```

error.go
rossf7: OK no worries. Thanks for getting this sorted.

PR #96, opened 2019-04-16
Parse yaml

rossf7:
Towards giantswarm/giantswarm#4949

This is prep for https://github.com/giantswarm/helmclient/pull/94. In chart-operator we need to merge configmap and secret values stored as YAML.

The merge logic needs the inputs to be `map[string]interface{}` types. But go-yaml V2 returns a `map[interface{}]interface{}` . They plan to change this in go-yaml V3.

This implements a workaround suggested in the upstream issue https://github.com/go-yaml/yaml/issues/139.

### DIFF
--- a/merge_values.go
+++ b/merge_values.go
@@ -1,0 +1,57 @@
+package helmclient
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// yamlToStringMap unmarshals the YAML input into a map[string]interface{}
+// with string keys. This is necessary because the default behaviour of the
+// YAML parser is to return map[interface{}]interface{} types.
+// See https://github.com/go-yaml/yaml/issues/139.
+//
+func yamlToStringMap(input []byte) (map[string]interface{}, error) {
+	var raw interface{}
+	var result map[string]interface{}
+
+	err := yaml.Unmarshal(input, &raw)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	output := processMapValue(raw)
+	result = output.(map[string]interface{})
+
+	return result, nil
+}
+
+func processInterfaceArray(in []interface{}) []interface{} {
+	res := make([]interface{}, len(in))
+	for i, v := range in {
+		res[i] = processMapValue(v)
+	}
+	return res
+}
+
+func processInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+	for k, v := range in {
+		res[fmt.Sprintf("%v", k)] = processMapValue(v)
+	}
+	return res
+}
+
+func processMapValue(v interface{}) interface{} {
+	switch v := v.(type) {
+	case []interface{}:
+		return processInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return processInterfaceMap(v)
+	case string:
+		return v
+	default:
+		return fmt.Sprintf("%v", v)
+	}
+}
--- a/merge_values.go
+++ b/merge_values.go
@@ -21,8 +21,12 @@ func yamlToStringMap(input []byte) (map[string]interface{}, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	output := processMapValue(raw)
-	result = output.(map[string]interface{})
+	inputMap, ok := raw.(map[interface{}]interface{})
+	if !ok {
+		return nil, microerror.Maskf(executionFailedError, "input type %T but expected %T", raw, inputMap)
+	}
+
+	result = processInterfaceMap(inputMap)
 
 	return result, nil
 }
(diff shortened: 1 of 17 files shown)


merge_values.go
```diff
@@ -0,0 +1,57 @@
...
+func processMapValue(v interface{}) interface{} {
+	switch v := v.(type) {
+	case []interface{}:
+		return processInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return processInterfaceMap(v)
+	case string:
+		return v
+	default:
+		return fmt.Sprintf("%v", v)
```
kopiczko: I believe this is safer to return error here.

merge_values.go
```diff
@@ -0,0 +1,57 @@
...
+	var raw interface{}
+	var result map[string]interface{}
+
+	err := yaml.Unmarshal(input, &raw)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	output := processMapValue(raw)
+	result = output.(map[string]interface{})
```
kopiczko: I don't think it's correct to always assert processMapValue output type as `map[string]interface{}`. Why it wouldn't return that type instead?

rossf7: I want to base the merge logic on the upstream logic in Helm. It does what we need and we already use it. Albeit indirectly because we shell out to Helm in draughtsman.

https://github.com/helm/helm/blob/240e539cec44e2b746b3541529d41f4ba01e77df/cmd/helm/install.go#L358

For the merge logic to work it needs string keys. This is so it can identify nested values and do the recursive merge.

merge_values.go
kopiczko: But isn't that incorrect? What if input is an int for example?

merge_values.go
kopiczko: Sure, what I'm trying to say here this type assertion will fail at runtime when input will be of type string for example.

Unless there is an assumption that the input is always an object. But then I think it should be clear to avoid confusion. Like I have right now. This could be achieved by extracting `processValue` from `processMapValue`. Then `processMapValue` can return the desired type.

merge_values.go
rossf7: We need to be able to pass int values. So maybe the logic needs to handle more types. I'll do some more testing with chart-operator.

merge_values.go
rossf7: OK I think I get you. Extracting `processValue` is a good shout. I'll look into that.

merge_values.go
kopiczko: I think it's better to return an error on unhandled type and add handling as we see the error than making :hankey: silently.

rossf7: Changes made. @kopiczko PTAL

merge_values.go
```diff
@@ -0,0 +1,63 @@
...
+	case int:
+		return v
+	case string:
+		return v
+	case []interface{}:
+		return processInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return processInterfaceMap(v)
+	default:
+		return microerror.Maskf(executionFailedError, "%#v with type %T not supported")
```
kopiczko:
```suggestion
		return microerror.Maskf(executionFailedError, "value %#v with type %T not supported", v, v)
```

merge_values.go
```diff
@@ -0,0 +1,63 @@
...
+
+func processInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+	for k, v := range in {
+		res[fmt.Sprintf("%v", k)] = processMapValue(v)
+	}
+	return res
+}
+
+func processMapValue(v interface{}) interface{} {
```
kopiczko: I think this should be called `processValue`.

merge_values.go
kopiczko: What I'd like to do here is something like:

```go
inputMap, ok := raw.(map[interface{}]interface{})
if !ok {
	return nil, return microerror.Maskf(executionFailedError, "input type %T but expected %T", raw, inputMap)
}

result = processInterfaceMap(raw)
```

This would make clear what is the expected input type.

Other question: why we can't just do following in the first place?

```go
var raw map[string]interface{}

err := yaml.Unmarshal(input, &raw)
```

rossf7: That is nice to check the input type. Done.

The problem is the nested keys will be `interface{}` and not `string`. In V3 of `go-yaml` they want to change the behaviour but there is no date for V3. Until then this workaround is needed.

merge_values.go
rossf7: Damn that was all kinds of wrong. Thanks, fixed!

rossf7:
@kopiczko PTAL. Should be good now.

merge_values.go
kopiczko: Ack 👍 Maybe a comment in the code?

rossf7: This is already in the godoc and it links to the upstream issue. Going with this.

PR #97, opened 2019-04-18
Add new description field for status update

tomahawk28:
Toward: https://github.com/giantswarm/giantswarm/issues/5751

After this, I would need to update `chartstatus` resource.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -460,6 +460,7 @@ func (c *Client) GetReleaseHistory(ctx context.Context, releaseName string) (*Re
 
 		history = &ReleaseHistory{
 			AppVersion:   appVersion,
+			Description:  release.Info.Description,
 			LastDeployed: lastDeployed,
 			Name:         release.Name,
 			Version:      version,
--- a/types.go
+++ b/types.go
@@ -22,6 +22,8 @@ type ReleaseContent struct {
 type ReleaseHistory struct {
 	// AppVersion is the app version of the Helm Chart that has been deployed.
 	AppVersion string
+	// Description is the incident log about past Helm chart deployment issue.
+	Description string
 	// LastDeployed is the time the Helm Chart was last deployed.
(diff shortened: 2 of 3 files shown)


helmclient.go
```diff
@@ -460,6 +460,7 @@ func (c *Client) GetReleaseHistory(ctx context.Context, releaseName string) (*Re
 
 		history = &ReleaseHistory{
 			AppVersion:   appVersion,
+			Description:  release.Info.Description,
```
rossf7: <3

types.go
```diff
@@ -22,6 +22,8 @@ type ReleaseContent struct {
 type ReleaseHistory struct {
 	// AppVersion is the app version of the Helm Chart that has been deployed.
 	AppVersion string
+	// Description is the incident log about past Helm chart deployment issue.
```
rossf7: Is incident log helm terminology? This isn't correct e.g. if the helm release has been upgraded the text will be `Upgrade complete`.

This should describe what information is provided by the tiller API. Not how we want to use it in the operator.

tomahawk28: ```	// Description is the log message about past Helm chart deployment provided by tiller API.``` WDYT?

types.go
```diff
@@ -22,6 +22,8 @@ type ReleaseContent struct {
 type ReleaseHistory struct {
 	// AppVersion is the app version of the Helm Chart that has been deployed.
 	AppVersion string
+	// Description is the log message about past Helm chart deployment provided by tiller API.
```
rossf7:
```suggestion
	// Description is a human-friendly "log entry" about this Helm release.
```

Let's use this based on what they have in upstream.

https://github.com/helm/helm/blob/ae17f3199e519126a83409d08d534d12dadd75aa/_proto/hapi/release/info.proto#L35

tomahawk28: Much simple 😆

PR #100, opened 2019-05-28
Helmclient metrics

tomahawk28:
- Adding error gauge, the time duration for each operation on helmclient.


helmclient.go
```diff
@@ -118,7 +119,9 @@ func New(config Config) (*Client, error) {
 }
 
 // DeleteRelease uninstalls a chart given its release name.
-func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options ...helmclient.DeleteOption) error {
+func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options ...helmclient.DeleteOption) (err error) {
```
tomahawk28: I named on returning `error` variable so we can check it whether it's nil or not when we exit this function.

helmclient.go
```diff
@@ -118,7 +119,9 @@ func New(config Config) (*Client, error) {
...
+	eventName := "delete_release"
```
tomahawk28: I'm a quite python fan, but not sure what is `__func__` equivalent for golang.

helmclient.go
```diff
@@ -574,85 +597,107 @@ func (c *Client) ListReleaseContents(ctx context.Context) ([]*ReleaseContent, er
...
 
-	resChan, errChan := c.newHelmClientFromTunnel(t).RunReleaseTest(releaseName, helmclient.ReleaseTestTimeout(int64(runReleaseTestTimout)))
+	resChan, errChan := c.newHelmClientFromTunnel(tunnel).RunReleaseTest(releaseName, helmclient.ReleaseTestTimeout(int64(runReleaseTestTimout)))
 	if IsReleaseNotFound(err) {
 		return backoff.Permanent(microerror.Mask(err))
 	} else if err != nil {
 		return microerror.Mask(err)
 	}
 
+W:
```
tomahawk28: Adding the label for exiting from the infinite loop.

helmclient.go
tomahawk28: Should it be const strings?

metric.go
```diff
@@ -0,0 +1,36 @@
+package helmclient
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	PrometheusNamespace = "operatorkit"
```
rossf7: Should this be `helmclient`?

helmclient.go
```diff
@@ -574,85 +597,107 @@ func (c *Client) ListReleaseContents(ctx context.Context) ([]*ReleaseContent, er
...
 	for {
 		select {
 		case err := <-errChan:
 			if err != nil {
-				return microerror.Mask(err)
+				err = microerror.Mask(err)
+				break W
```
rossf7: Why is this `break W`?

helmclient.go
rossf7: I'm not sure either. Maybe the `reflect` package? I'd suggest researching it if you want to use this.

helmclient.go
kopiczko: I still don't get why. AFAICT this should be used only for documentation purposes.

I also don't think `errorReporting` is reliable. I already see some errors redeclared in backoff functions for example.

I'd rather go with `deleteRelease` called by `DeleteRelease` and check for errors there.

metric.go
tomahawk28: 🤦‍♂ Oh, yes.

helmclient.go
tomahawk28: Otherwise, it would only break inner `select` statement.

helmclient.go
```diff
@@ -361,6 +361,14 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 // values provided when the chart was installed. The releaseName is the name
 // of the Helm Release that is set when the Helm Chart is installed.
 func (c *Client) GetReleaseContent(ctx context.Context, releaseName string) (*ReleaseContent, error) {
+	releaseContent, err := c.getReleaseContent(ctx, releaseName)
```
tomahawk28: I follow Pawel's suggestion to check the error result and update metrics.

Gopkg.toml
```diff
@@ -107,3 +107,7 @@ required = [
 [prune]
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "github.com/prometheus/client_golang"
+  version = "0.9.3"
```
rossf7: This is in the wrong place.

helmclient.go
```diff
@@ -361,6 +362,17 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 // values provided when the chart was installed. The releaseName is the name
 // of the Helm Release that is set when the Helm Chart is installed.
 func (c *Client) GetReleaseContent(ctx context.Context, releaseName string) (*ReleaseContent, error) {
+	eventName := "get_release_content"
```
rossf7: I find this hard to read. Please split up the lines with whitespace.

helmclient.go
```diff
@@ -3,6 +3,7 @@ package helmclient
 import (
 	"context"
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus"
```
rossf7: Needs ordering.

metric.go
```diff
@@ -0,0 +1,36 @@
...
+			Help:      "Number of helmclient errors.",
+		},
+		[]string{"event"},
+	)
+	controllerHistogram = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: PrometheusNamespace,
+			Subsystem: PrometheusSubsystem,
+			Name:      "event",
+			Help:      "Histogram for events within the helmclient controller.",
```
rossf7:
```suggestion
			Help:      "Histogram for events within the helmclient library.",
```

helmclient.go
```diff
@@ -361,6 +362,21 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 // values provided when the chart was installed. The releaseName is the name
 // of the Helm Release that is set when the Helm Chart is installed.
 func (c *Client) GetReleaseContent(ctx context.Context, releaseName string) (*ReleaseContent, error) {
+	eventName := "get_release_content"
+
+	t := prometheus.NewTimer(controllerHistogram.WithLabelValues(eventName))
+
+	releaseContent, err := c.getReleaseContent(ctx, releaseName)
+	if err != nil {
+		controllerErrorGauge.WithLabelValues(eventName).Inc()
```
rossf7: I think the error masking won't work here and then the stack traces will be wrong. 

After updating the gauge I think we should return the masked error like we usually do.

helmclient.go
```diff
@@ -361,6 +362,21 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
+	}
+
+	t.ObserveDuration()
+
+	return releaseContent, err
```
rossf7:
```suggestion
	return releaseContent, nil
```

helmclient.go
tomahawk28: Good catch, never thought about that.

helmclient.go
```diff
@@ -361,6 +362,22 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
 // of the Helm Release that is set when the Helm Chart is installed.
 func (c *Client) GetReleaseContent(ctx context.Context, releaseName string) (*ReleaseContent, error) {
+	eventName := "get_release_content"
+
+	t := prometheus.NewTimer(histogram.WithLabelValues(eventName))
+
+	releaseContent, err := c.getReleaseContent(ctx, releaseName)
+	if err != nil {
+		errorGauge.WithLabelValues(eventName).Inc()
+		return releaseContent, microerror.Mask(err)
```
rossf7:
```suggestion
		return nil, microerror.Mask(err)
```

helmclient.go
```diff
@@ -361,6 +362,22 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
+	}
+
+	t.ObserveDuration()
```
JosephSalisbury: Why not defer? Then you can keep it closer to the timer creation

PR #101, opened 2019-05-29
Support merging and processing null valued keys

sslavic:
Setting leaf or even higher level key to `null` is official way of overriding default key setting with intent to having it deleted (see [docs](https://helm.sh/docs/chart_template_guide/#deleting-a-default-key)). To be able to use `helmclient.MergeValues` in `opsctl ensure appcatalogs` command logic and have the command properly supporting customization / overrides / key deletion via `null` value, merging and processing `null` valued keys needs to be supported. This PR adds that feature.

### DIFF
--- a/merge_values.go
+++ b/merge_values.go
@@ -135,6 +135,12 @@ func processInterfaceMap(in map[interface{}]interface{}) map[string]interface{} 
 }
 
 func processValue(v interface{}) interface{} {
+	// yaml null-valued key unmarshalls to nil value in Go, so nil is a valid value that has to be handled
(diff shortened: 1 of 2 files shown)


merge_values.go
```diff
@@ -135,6 +135,10 @@ func processInterfaceMap(in map[interface{}]interface{}) map[string]interface{}
 }
 
 func processValue(v interface{}) interface{} {
+	if v == nil {
```
rossf7: What about linking to the upstream docs here?

sslavic: Done

PR #106, opened 2019-06-20
Add network policy for tiller

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4033

Code generates policy:

```
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  creationTimestamp: "2019-06-19T17:06:25Z"
  generation: 2
  name: tiller-deploy
  namespace: giantswarm
  resourceVersion: "1170583"
  selfLink: /apis/extensions/v1beta1/namespaces/giantswarm/networkpolicies/tiller-deploy
  uid: 9258a16b-92b4-11e9-b419-000d3a44433d
spec:
  egress:
  - to:
    - ipBlock:
        cidr: 10.0.0.0/8
    - ipBlock:
        cidr: 172.16.0.0/12
    - ipBlock:
        cidr: 192.168.0.0/16
  ingress:
  - ports:
    - port: 44134
      protocol: TCP
    - port: 44135
      protocol: TCP
  podSelector:
    matchLabels:
      app: helm
      name: tiller
  policyTypes:
  - Egress
  - Ingress
```

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -9,9 +9,11 @@ import (
 	"github.com/giantswarm/errors/guest"
 	"github.com/giantswarm/microerror"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/helm/cmd/helm/installer"
 )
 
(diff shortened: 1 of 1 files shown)


ensure_tiller_installed.go
```diff
@@ -120,6 +122,91 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
+							},
+						},
+					},
+				},
+				Egress: []networkingv1.NetworkPolicyEgressRule{
+					networkingv1.NetworkPolicyEgressRule{
+						To: []networkingv1.NetworkPolicyPeer{
+							networkingv1.NetworkPolicyPeer{
+								IPBlock: &networkingv1.IPBlock{
+									CIDR: "10.0.0.0/8",
```
pipo02mix: Is this the range for the nodes?
IIRC this tiller is installed in the tenant clusters and we have some clusters with different node IP ranges. Am I right?

rossf7: I'm not sure on the CIDRs but yes this is used to install tiller in control plane and tenant clusters. So we need to be able to support both.

corest: Those are reanges of private networks. It is not possible to restrict access in super granular way as we just don't have data about host networks and pod cidrs.

Anyway, this list should be changed to allow all the traffic as I've tested tiller after all the components were installed so I didn't catch external requests. Next commit fixes that

xh3b4sd:
Thank you for your contribution here. I just want to say that the design of our helm client is very unfortunate and that we have to adapt the implementation in a more proper way so that we can manage reconciliation in a more reliable and non blocking way. The code here is only getting bigger and unmanageable. All of this adds up and harms seamless operations.

rossf7:
> the design of our helm client is very unfortunate

@xh3b4sd Totally agree and yes especially `EnsureTillerInstalled` has gotten too complicated. Creating the `tiller` resource has helped and we could make it more like a regular `operatorkit` resource and remove from `helmclient`.

The problem is I just don't want to invest lots of time here. We will remove Tiller from our architecture as soon as we have a viable way of doing it.

Helm 3 is in alpha and I tried it last week. It's early but I like the direction. They are aiming to go GA in September but I'm skeptical about this date. There are other ways of removing Tiller but my preference is to adopt Helm 3 if it works for us. As I want to stay on "mainline Helm" if we can.

cc @giantswarm/team-batman

corest:
Tested with draughtsman - works fine. Will merge later today

PR #107, opened 2019-06-24
K8s 1.15

tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/6302

PR #109, opened 2019-06-25
use new k8sclient lib

xh3b4sd:
Based on https://github.com/giantswarm/helmclient/pull/107. Aim here is to decouple `helmclient` from `e2esetup`. Once #107 is merged we should try to get this sorted. Right now I couldn't make it work due to `dep` magic. I think the helm deps are still weird.


Gopkg.toml
```diff
@@ -47,8 +47,8 @@ required = [
   name = "github.com/giantswarm/backoff"
 
 [[constraint]]
-  branch = "k8s-1.15"
-  name = "github.com/giantswarm/e2esetup"
+  branch = "master"
+  name = "github.com/giantswarm/clientk8s"
```
kopiczko: Why this repo has inverted name? So far we had `helmclient` and not `clienthelm`.

xh3b4sd: Because this is what Theo suggested and I did not mind either way.

PR #112, opened 2019-07-25
Make max wait for EnsureTillerInstalled configurable

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6561

This is the other change needed. It improves the backoff handling and makes the max wait configurable. We keep the default 60 seconds for e2e tests.

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -191,12 +191,13 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 	var err error
 	var installTiller bool
+	var maxWait time.Duration
 	var pod *corev1.Pod
 	var upgradeTiller bool
 
 	{
 		o := func() error {
-			pod, err = getPod(c.k8sClient, tillerLabelSelector, c.tillerNamespace)
+			pod, err = getPod(c.k8sClient, c.tillerNamespace)
 			if IsNotFound(err) {
 				// Fall through as we need to install Tiller.
 				installTiller = true
@@ -207,10 +208,18 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 			return nil
 		}
-		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
+
+		// maxWait is configurable so we can use more time in e2e tests
+		// than in operators.
+		maxWait, err = time.ParseDuration(c.ensureTillerInstalledMaxWait)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		b := backoff.NewConstant(maxWait, 3*time.Second)
 		n := backoff.NewNotifier(c.logger, context.Background())
 
-		err := backoff.RetryNotify(o, b, n)
+		err = backoff.RetryNotify(o, b, n)
 		if err != nil {
 			return microerror.Mask(err)
 		}
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -191,7 +191,6 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 	var err error
 	var installTiller bool
-	var maxWait time.Duration
 	var pod *corev1.Pod
 	var upgradeTiller bool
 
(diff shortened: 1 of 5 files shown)


ensure_tiller_installed.go
```diff
@@ -207,10 +208,18 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 			return nil
 		}
-		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
+
+		// maxWait is configurable so we can use more time in e2e tests
+		// than in operators.
```
rossf7: The current value is 60 seconds which I set because we install tiller in e2e tests and in operators.

But 60 seconds is too long for cluster-operator so I'm making it configurable.

ensure_tiller_installed.go
```diff
@@ -207,10 +208,18 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
+		maxWait, err = time.ParseDuration(c.ensureTillerInstalledMaxWait)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		b := backoff.NewConstant(maxWait, 3*time.Second)
```
rossf7: I think constant with a shorter interval makes more sense.

helmclient_test.go
```diff
@@ -488,7 +488,7 @@ func Test_Client_ListReleaseContents(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(result, tc.expectedContents) {
-				t.Fatalf("Releases == %q, want %q", result, tc.expectedContents)
+				t.Fatalf("Releases == %#v, want %#v", result, tc.expectedContents)
```
rossf7: This was wrong and in Go 1.12 was causing the test to fail.

helmclient.go
```diff
@@ -721,9 +726,10 @@ func filterList(rels []*hapirelease.Release) []*hapirelease.Release {
 	return uniq
 }
 
-func getPod(client kubernetes.Interface, labelSelector, namespace string) (*corev1.Pod, error) {
+func getPod(client kubernetes.Interface, namespace string) (*corev1.Pod, error) {
 	o := metav1.ListOptions{
-		LabelSelector: labelSelector,
+		FieldSelector: tillerFieldSelector,
+		LabelSelector: tillerLabelSelector,
```
rossf7: We now also check that the tiller pod is running. We need to make a port forwarding connection so if its not running there is no point trying.

error.go
```diff
@@ -186,15 +186,6 @@ func IsTestReleaseTimeout(err error) bool {
...
 }
 
-var tillerInstallationFailedError = &microerror.Error{
-	Kind: "tillerInstallationFailedError",
-}
-
-// IsTillerInstallationFailed asserts tillerInstallationFailedError.
-func IsTillerInstallationFailed(err error) bool {
-	return microerror.Cause(err) == tillerInstallationFailedError
-}
```
rossf7: So this error isn't used anywhere but its the error we match on in cluster-operator. 

picard-facepalm

ensure_tiller_installed.go
```diff
@@ -207,10 +208,18 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 			return nil
 		}
-		b := backoff.NewExponential(1*time.Minute, 5*time.Second)
+
+		// maxWait is configurable so we can use more time in e2e tests
+		// than in operators.
+		maxWait, err = time.ParseDuration(c.ensureTillerInstalledMaxWait)
```
xh3b4sd: Considering this here is a library we can also configure the properly typed `time.Duration`. No need for the string magic and runtime parsing I think.

rossf7: It's dumb but I had trouble comparing the `time.Duration` to see if its set. If it's unset we need to use the 60 second default.

So that's why I did it this way. Any ideas?

rossf7: OK its 0 I'l get it fixed 😄

rossf7: Done and it's way nicer.

kopiczko:
I think we should keep the default for production use case and change to something else in e2e tests. Why we can have the same for both though and just increase the timeout?

kopiczko:
I think we should keep the default for production use case and change to something else in e2e tests. Why we can have the same for both though and just increase the timeout?

rossf7:
If we change it in the e2e tests that is a lot more places we need to update. I don't want to invest lots of time here because we will throw away Tiller.

I'll configure the shorter max wait in cluster-operator, chart-operator and draughtsman which is where the longer max wait hurts us.

kopiczko:
Throwing away tiller is convincing. I hope that'll happen sooner than later :D

rossf7:
The Helm team are aiming to ship Helm 3 in September at Helm Summit. The alpha 2 was released this week.

https://github.com/helm/helm/releases/tag/v3.0.0-alpha.2

If it slips we may have to rethink but they are dropping features to make the deadline which is a good call IMO. cc @giantswarm/team-batman

PR #115, opened 2019-07-31
helmclient: don't print response body in error for 404

kopiczko:


pull_chart_tarball.go
```diff
@@ -47,7 +47,12 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 			if err != nil {
 				return microerror.Mask(err)
 			}
-			return microerror.Maskf(executionFailedError, fmt.Sprintf("got StatusCode %d with body %s", resp.StatusCode, buf.String()))
+			// Github pages 404 produces full HTML page which
+			// obscures the logs.
+			if resp.StatusCode == http.StatusNotFound {
```
TheoBrigitte: I don't see why we need a special case for 404, printing status code, url and body every time sounds good to me. No blocker.

kopiczko: It outputs full html github page when it's 404 which is useless. I don't know what happens for other statuses.

PR #116, opened 2019-08-09
Add podsecuritypolicy for Tiller

glitchcrab:
This PR adds a clusterrole, clusterrolebinding and a podsecuritypolicy for Tiller. 

My Go experience is relatively limited (especially so when using client_go) so please expect to have to point me in the right direction - I wouldn't expect this to be the finished article. It is currently untested as I wanted to get a first-pass from someone with more knowledge than myself. 

Provided the resources are created correctly, Tiller should continue to work fine with the PSP in place. It currently runs as `nobody/nogroup` and only requires access to secrets.


glitchcrab:
superseded by a newer PR which actually works

PR #120, opened 2019-09-24
Tiller PSP

glitchcrab:
towards giantswarm/giantswarm/issues/6013

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -126,10 +126,10 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 	{
 		name := fmt.Sprintf("%s-psp", tillerPodName)
 
-		// apiGroups     := []string{"extensions"}
-		// resources     := []string{"podsecuritypolicies"}
-		// resourceNames := []string{name}
-		// verbs         := []string{"use"}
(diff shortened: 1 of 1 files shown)


glitchcrab:
Well this is less than ideal - I didn't think about forcing Tiller to run as non-root (which it can do, I've tested it manually). It doesn't look like the Helm installer exposes the `securityContext` either: https://github.com/helm/helm/blob/master/cmd/helm/installer/install.go#L200

Since deploying in a TC I've noticed that the PSP needs a little tweaking anyway, but I presume it's going to be a lot of work to make it run as non-root, especially given that we will hopefully do away with Tiller in the not-too-distant future?

I'd appreciate your thoughts @giantswarm/team-batman

rossf7:
Yes the Helm installer is very basic and misses settings. This also hurts us by preventing us from setting TILLER_MAX_HISTORY on existing clusters.

From my POV I'd prefer if we didn't run Tiller as non-root if it adds complexity and since we will remove Tiller. Adding the PSP via helmclient is already non-trivial.

But I don't know how important it is that Tiller runs as non-root. @puja108 @corest WDYT?

rossf7:
@glitchcrab Oh wow I had an open tab that hadn't updated 🤦‍♂ 

Anyway that was my 2 cents! 😆

PR #124, opened 2019-10-02
re-add PSP and RBAC

glitchcrab:
This reverts commit 38a0673024661c036251904c7b2ed8a0462fb771.

This was originally reverted to avoid multiple changes being vendored into various operators at once.


glitchcrab:
@rossf7 can you see any problems with merging this again?

rossf7:
@glitchcrab It should be OK and for the moment we're not making many helmclient changes.

It just needs careful testing and I would test it in both CP and TC. As in CP both draughtsman and chart-operator are using tiller.

glitchcrab:
@rossf7 sure, this was merged before (but then reverted) so it's already had some testing. My plan was to get this merged and then work through the repos which vendor it to test them one at a time

PR #128, opened 2019-10-24
Return permanent error in case of tarball not found

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7359

Return `Permanent` error in case of no host issue or HTTP not found.

### DIFF
--- a/error.go
+++ b/error.go
@@ -87,6 +87,27 @@ func IsNotFound(err error) bool {
 }
 
 var (
+	noHostErrorString = "no such host"
+)
+
+var noHostError = &microerror.Error{
+	Kind: "noHostError",
+}
+
+// IsNoHostError asserts noHostError.
+func IsNoHostError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if err == noHostError {
+		return true
+	}
+
+	return strings.Contains(err.Error(), noHostErrorString)
+}
+
+var (
 	releaseAlreadyExistsRegexp = regexp.MustCompile(`release named \S+ already exists`)
 )
 
--- a/error.go
+++ b/error.go
@@ -87,27 +87,6 @@ func IsNotFound(err error) bool {
 }
 
 var (
-	noHostErrorString = "no such host"
-)
-
-var noHostError = &microerror.Error{
-	Kind: "noHostError",
-}
-
-// IsNoHostError asserts noHostError.
-func IsNoHostError(err error) bool {
-	if err == nil {
-		return false
-	}
-
-	if err == noHostError {
-		return true
-	}
-
-	return strings.Contains(err.Error(), noHostErrorString)
-}
-
-var (
 	releaseAlreadyExistsRegexp = regexp.MustCompile(`release named \S+ already exists`)
 )
 
--- a/error_internal.go
+++ b/error_internal.go
@@ -1,0 +1,16 @@
+package helmclient
+
+import "strings"
+
+var (
+	noHostErrorString = "no such host"
+)
+
+// IsNoHostError asserts noHostError.
+func IsNoHostError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return strings.Contains(err.Error(), noHostErrorString)
+}
--- a/error_internal.go
+++ b/error_internal.go
@@ -2,7 +2,7 @@ package helmclient
 
 import "strings"
 
-// isNoHostError asserts no route to Host error.
+// isNoSuchHostError asserts no such host error.
 func isNoHostError(err error) bool {
 	if err == nil {
 		return false
--- a/error_internal.go
+++ b/error_internal.go
@@ -2,15 +2,11 @@ package helmclient
 
 import "strings"
 
-var (
-	noHostErrorString = "no such host"
-)
-
-// IsNoHostError asserts noHostError.
-func IsNoHostError(err error) bool {
+// isNoHostError asserts no route to Host error.
+func isNoHostError(err error) bool {
 	if err == nil {
 		return false
 	}
 
-	return strings.Contains(err.Error(), noHostErrorString)
+	return strings.Contains(err.Error(), "no such host")
 }
--- a/error_internal.go
+++ b/error_internal.go
@@ -3,7 +3,7 @@ package helmclient
 import "strings"
 
 // isNoSuchHostError asserts no such host error.
-func isNoHostError(err error) bool {
+func isNoSuchHostError(err error) bool {
 	if err == nil {
 		return false
 	}
--- a/pull_chart_tarball.go
+++ b/pull_chart_tarball.go
@@ -36,10 +36,9 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error) 
 
 	o := func() error {
 		resp, err := c.httpClient.Do(req)
-		if err != nil {
-			if IsNoHostError(err) {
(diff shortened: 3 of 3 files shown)


pull_chart_tarball.go
```diff
@@ -37,7 +37,7 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 	o := func() error {
 		resp, err := c.httpClient.Do(req)
 		if err != nil {
-			return microerror.Mask(err)
+			return backoff.Permanent(microerror.Mask(err))
```
tomahawk28: Cancelling retry if there is no such host. 
e.g. 
```
W 10/24 10:07:06 /apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/charts/bad-url releasev1.GetDesiredState retrying due to error | operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:96 | controller=chart-operator-chart | event=update | loop=3 | underlyingResource=releasev1 | version=1305
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:89:
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/service/controller/chart/v1/resource/release/desired.go:30:
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/pull_chart_tarball.go:26:
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/pull_chart_tarball.go:79:
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/backoff/retry.go:23:
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/pull_chart_tarball.go:40:
	Get https://just-a-bad-url.nothing/non-existent.tgz: dial tcp: lookup just-a-bad-url.nothing: no such host
```

tomahawk28: Would it make sense just grep the error string to find out whether this is no host issue? I think network error types vary so it doesn't matter to retry at this moment.

rossf7: Yes I think this is too generic as is.

What about an internal error matcher for no such host? So we only use backoff.Permanent then.

pull_chart_tarball.go
```diff
@@ -50,7 +50,7 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 			// Github pages 404 produces full HTML page which
 			// obscures the logs.
 			if resp.StatusCode == http.StatusNotFound {
-				return microerror.Maskf(executionFailedError, fmt.Sprintf("got StatusCode %d for url %#q", resp.StatusCode, req.URL.String()))
+				return backoff.Permanent(microerror.Maskf(executionFailedError, fmt.Sprintf("got StatusCode %d for url %#q", resp.StatusCode, req.URL.String())))
```
rossf7: This makes total sense and I think a 404 is more likely. As it can happen if you choose the wrong version. 

For the host we always use GitHub at the moment.

tomahawk28:
Tested on a kind cluster, both `404` and no route cases. PTAL.

error.go
```diff
@@ -86,6 +86,27 @@ func IsNotFound(err error) bool {
...
+	if err == nil {
+		return false
+	}
+
+	if err == noHostError {
+		return true
+	}
+
+	return strings.Contains(err.Error(), noHostErrorString)
+}
```
kopiczko: This should go to `error_internal.go` as `isNoHostError` w/o corresponding `noHosError` variable. See https://github.com/giantswarm/fmt/blob/master/go/errors.md#handling-third-party-library-errors

error_internal.go
```diff
@@ -0,0 +1,16 @@
+package helmclient
+
+import "strings"
+
+var (
+	noHostErrorString = "no such host"
```
rossf7:
```suggestion
	noSuchHostErrorString = "no such host"
```

error_internal.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// IsNoHostError asserts noHostError.
+func IsNoHostError(err error) bool {
```
rossf7:
```suggestion
func isNoHostError(err error) bool {
```

It's an internal error matcher so it shouldn't be exported.

pull_chart_tarball.go
```diff
@@ -37,6 +37,9 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 	o := func() error {
 		resp, err := c.httpClient.Do(req)
 		if err != nil {
+			if IsNoHostError(err) {
```
rossf7: Why the nested if here?

error_internal.go
```diff
@@ -0,0 +1,16 @@
...
+	if err == nil {
+		return false
+	}
+
+	return strings.Contains(err.Error(), noHostErrorString)
```
kopiczko: I don't see a point of having `noHostErrorString` as a const.

tomahawk28:
I saw another error with a different message. 
```
W 10/24 15:20:24 /apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/charts/bad-url releasev1.GetDesiredState retrying backoff in '5s' due to error | backoff/notifier.go:13 | controller=chart-operator-chart | event=update | loop=14 | version=19700
	/Users/jihyukbok/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/pull_chart_tarball.go:42:
	Get https://giantswarm.kithub.com/control-plane-test-catalog/aws-operator-1.0.0-33094d81c676f074856c831514aceb09f1a9458bb.tgz: dial tcp 64.98.145.30:443: connect: connection refused
```
As I said, it could be varied in network errors. So I would reduce the retry as 3 times in helmclient, and will cancel the resource in `chart-operator` if it's failed at first try. WDYT?

tomahawk28:
Tested again on KIND cluster. Works well. 
@rossf7, @kopiczko PTAL

error_internal.go
```diff
@@ -0,0 +1,12 @@
+package helmclient
+
+import "strings"
+
+// isNoHostError asserts no route to Host error.
```
rossf7:
```suggestion
// isNoSuchHostError asserts no such host error.
```

The error message text "no such host" is descriptive for me. So let's align our code with that.

error_internal.go
```diff
@@ -0,0 +1,12 @@
...
+func isNoHostError(err error) bool {
```
rossf7:
```suggestion
func isNoSuchHostError(err error) bool {
```

pull_chart_tarball.go
```diff
@@ -36,7 +36,9 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 
 	o := func() error {
 		resp, err := c.httpClient.Do(req)
-		if err != nil {
+		if isNoHostError(err) {
+			return backoff.Permanent(microerror.Maskf(executionFailedError, "couldn't access to host %#v", req.Host))
```
rossf7:
```suggestion
			return backoff.Permanent(microerror.Maskf(executionFailedError, "no such host %#q", req.Host))
```

PR #129, opened 2019-10-25
Return pull chart error type on when pulling failed

tomahawk28:
Following https://github.com/giantswarm/chart-operator/pull/308#discussion_r338917401

> Is there a more specific error that we can match against? My thinking is that there is a specific case in which we really want to cancel. On the other hand there are cases which we do not expect and we should treat them like real errors.

### DIFF
--- a/pull_chart_tarball.go
+++ b/pull_chart_tarball.go
@@ -37,7 +37,7 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error) 
 	o := func() error {
 		resp, err := c.httpClient.Do(req)
 		if isNoSuchHostError(err) {
-			return backoff.Permanent(microerror.Maskf(executionFailedError, "no such host %#q", req.Host))
+			return backoff.Permanent(microerror.Maskf(pullChartFailedError, "no such host %#q", req.Host))
 		} else if err != nil {
 			return microerror.Mask(err)
 		}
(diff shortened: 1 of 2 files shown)


pull_chart_tarball.go
```diff
@@ -52,7 +52,7 @@ func (c *Client) doFile(ctx context.Context, req *http.Request) (string, error)
 			// Github pages 404 produces full HTML page which
 			// obscures the logs.
 			if resp.StatusCode == http.StatusNotFound {
-				return backoff.Permanent(microerror.Maskf(executionFailedError, fmt.Sprintf("got StatusCode %d for url %#q", resp.StatusCode, req.URL.String())))
+				return backoff.Permanent(microerror.Maskf(pullChartFailedError, fmt.Sprintf("got StatusCode %d for url %#q", resp.StatusCode, req.URL.String())))
```
xh3b4sd: Why do we want to treat these two cases the same way? Isn't there any value in differentiating them? "Not Found" is a pretty different root cause than "No Such Host". Why ruling them all the same way?

tomahawk28: There's no serious reason against differentiating it, the only reason was that I don't want to have multiple `if else` statement in `chart-operator` when matcher function in using.

tomahawk28: I would return `notFound` error in this case and return `pullChartURLError` for other cases, WDYT?

rossf7: I'm fine with `pullChartFailedError` for both.

We need to pull the chart tarball over HTTP and it doesn't work. That is what we care about. The reason why is clear from the error text.

tomahawk28:
an Issue created for CI: https://github.com/giantswarm/giantswarm/issues/7390

PR #130, opened 2019-10-25
Waiting for running tiller pod in e2e

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7390

Tiller pod could be temporarily in outage after changing a tiller version. 
We cannot assume that we need to install a fresh tiller rather than upgrade by just checking once.

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -197,11 +197,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 	{
 		o := func() error {
 			pod, err = getPod(c.k8sClient, c.tillerNamespace)
-			if IsNotFound(err) {
-				// Fall through as we need to install Tiller.
-				installTiller = true
-				return nil
-			} else if err != nil {
+			if err != nil {
 				return microerror.Mask(err)
 			}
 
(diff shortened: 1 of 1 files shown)


tomahawk28:
Tested twice, no error came out.

ensure_tiller_installed.go
```diff
@@ -197,11 +197,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 	{
 		o := func() error {
 			pod, err = getPod(c.k8sClient, c.tillerNamespace)
-			if IsNotFound(err) {
```
tomahawk28: Moving `IsNotFound` matcher usage outside of retries.

tomahawk28:
I'm really not sure what makes failed at this test, https://circleci.com/gh/giantswarm/helmclient/1960?pipelines-ui-opt-out

tfussell:
I see. All I can think of is adding a debug print statement to see what type the error is if it's not `IsNotFound` here https://github.com/giantswarm/helmclient/blob/f1d5a75b7dfc37edd4ca4f15de33a0368ea3d612/ensure_tiller_installed.go#L211

tfussell:
The second build with extra logging seems different from the first even though the code didn't change. Can we run this locally to debug?

tomahawk28:
It stranges, logs reports that tiller pod is even up-to-dated. Not sure what is broken.

tomahawk28:
closing in favor of https://github.com/giantswarm/helmclient/pull/131

PR #131, opened 2019-10-26
Fixing e2e error cases

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7390

Deleting preinstall procedure for `UpgradeTillerTest` and move it circleCI config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,15 @@ e2eTest: &e2eTest
           chmod +x ./kubectl
 
     - run:
+        name: Get helm binary
+        command: |
+          curl -LO https://get.helm.sh/helm-v2.14.2-linux-amd64.tar.gz
+          tar xzvf helm-v2.8.2-linux-amd64.tar.gz
+          mv linux-amd64/helm .
+          chmod +x helm
+          rm -rf linux-amd64/
+
+    - run:
         name: Init cluster
         command: |
           ./e2ectl cluster create
@@ -28,7 +37,7 @@ e2eTest: &e2eTest
           then
             export KUBECONFIG=$(./e2ectl kubeconfig path)
             ./kubectl create ns giantswarm
-            helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.12.0 --wait
+            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.12.0 --wait
           fi
 
     - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,17 +16,14 @@ e2eTest: &e2eTest
     - run:
         name: Get kubectl binary
         command: |
-          curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
           chmod +x ./kubectl
 
     - run:
         name: Get helm binary
         command: |
-          curl -LO https://get.helm.sh/helm-v2.12.0-linux-amd64.tar.gz
-          tar xzvf helm-v2.12.0-linux-amd64.tar.gz
-          mv linux-amd64/helm .
+          curl -L https://get.helm.sh/helm-v2.12.0-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
           chmod +x helm
-          rm -rf linux-amd64/
 
     - run:
         name: Init cluster
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ e2eTest: &e2eTest
         name: Get helm binary
         command: |
           curl -LO https://get.helm.sh/helm-v2.14.2-linux-amd64.tar.gz
-          tar xzvf helm-v2.8.2-linux-amd64.tar.gz
+          tar xzvf helm-v2.14.2-linux-amd64.tar.gz
           mv linux-amd64/helm .
           chmod +x helm
           rm -rf linux-amd64/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,13 +14,20 @@ e2eTest: &e2eTest
           chmod +x ./e2ectl
 
     - run:
+        name: Get kubectl binary
+        command: |
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+          chmod +x ./kubectl
+
+    - run:
         name: Init cluster
         command: |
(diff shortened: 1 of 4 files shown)


tomahawk28:
Tested 5 times, works fine.

.circleci/config.yml
```diff
@@ -13,11 +13,32 @@ e2eTest: &e2eTest
...
+          rm -rf linux-amd64/
+
     - run:
         name: Init cluster
         command: |
           ./e2ectl cluster create
           cp $(./e2ectl kubeconfig path) ${E2E_TEST_DIR}/kubeconfig
+          if [ -n "$HELM_INIT" ]
+          then
+            export KUBECONFIG=$(./e2ectl kubeconfig path)
```
tomahawk28: Instead of creating helm by codes in e2etest, it's better to install helm from outside of it.

ensure_tiller_installed.go
```diff
@@ -261,9 +261,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 		o := func() error {
 			t, err := c.newTunnel()
-			if !installTiller && IsTillerNotFound(err) {
-				return backoff.Permanent(microerror.Mask(err))
```
tomahawk28: Sometimes when the updated tiller status is `Pending` only, it would break the retry loop and e2etest would fail. It's worth to retry in this case rather than simply returning.

integration/test/upgradetiller/upgrade_tiller_test.go
```diff
@@ -16,46 +16,11 @@ import (
...
 
-	var err error
-
 	currentTillerImage := "quay.io/giantswarm/tiller:v2.14.3"
-	outdatedTillerImage := "quay.io/giantswarm/tiller:v2.12.0"
 
 	labelSelector := "app=helm,name=tiller"
 	tillerNamespace := "giantswarm"
 
-	// Install tiller using current image.
```
tomahawk28: Deleted the first `EnsureTillerInstalled` usage in `UpgradeTillerTest` since it's not what we want to test but too error-prone.

TheoBrigitte: Why are we dropping the downgrade part of the test ?

tomahawk28: We used `EnsureCreated` to create the latest version of tiller pod and downgrade it, but it was too error-prone; sometimes `getPod` didn't get the pod right for various reasons. I think installing helm from the beginning of e2etest is better to fit in real cases.

rossf7: I'm fine with the change of approach. The original approach I used was to install helm via EnsureTillerDeploying using the installer and then update the deployment to reduce the version.

Looks like this is unreliable and installing helm via the CLI is arguably a better test.

ensure_tiller_installed.go
rossf7: We need to be careful when changing this code and I'd prefer not to make major changes. 

We will likely need to upgrade to helm 2.15.0 but after that we may be able to move to Helm 3. The real fix here is to remove tiller. 😄

tomahawk28: I saw sometimes it also failed at `tooManyResultsError` since multiple tiller pods would run at the same time. 
https://github.com/giantswarm/helmclient/blob/2734fbbea648359692b6a6d3e8704026787cb053/helmclient.go#L738-L740

Do you think make sense to change like below, 
```golang
			if IsTillerNotFound(err) || IsTooManyResults(err) {
				return microerror.Mask(err)
			} else if err != nil {
			}
```

rossf7: However I'm not against this change if it is well tested. 

Removing the permanent backoff may make sense but IIRC max wait is 1 minute. So we need to make sure this doesn't slow down the tiller resources in the operators.

tomahawk28: @rossf7 I compared the time amount that circleCI took a week ago for `helmclient`, and it looks like doesn't have much difference. 
Moreover, those `if` statement would only executed when it's upgrading, so I think we are safe from removing `backoff.Permanent`

ensure_tiller_installed.go
```diff
@@ -261,10 +261,10 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
...
-			} else if err != nil {
+			if IsTillerNotFound(err) || IsTooManyResults(err) {
 				return microerror.Mask(err)
+			} else if err != nil {
+				return backoff.Permanent(microerror.Mask(err))
```
rossf7: AIUI the else is now a permanent backoff? But can there be cases we've missed?

If so we may not retry in situations where we currently do. That would be bad if we require another reconciliation loop to install tiller in tenant clusters.

ensure_tiller_installed.go
rossf7: @tomahawk28  My concern is how it behaves when vendored in app-operator. 

Please also test it manually so we're sure tiller can still be installed in tenant clusters OK.

ensure_tiller_installed.go
tomahawk28: @rossf7 By checking 3 all possible outcomes from tiller deployment phase (Tiller not found, multiple Tiller pods, version not matching), I think these are enough. 
And about stopping reconciliation of this  `backoff.Permanent` error cases, I think that depends on the behavior of caller, right? 
Do you think to make sense have another error, something like `ErrorStartingiTiller` as mask, and return in this case, so we can cancel our reconciliation?

rossf7: The approach makes sense. I'm fine with this once e2e is green and it has been tested manually.

tomahawk28:
Tested on `gorgoth`. I will create a following PR on `app-operator` to cancel the reconciliation in case of `tillerStartingError`.

tomahawk28:
@rossf7 @TheoBrigitte PTAL

ensure_tiller_installed.go
```diff
@@ -261,10 +261,10 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 		o := func() error {
 			t, err := c.newTunnel()
-			if !installTiller && IsTillerNotFound(err) {
-				return backoff.Permanent(microerror.Mask(err))
-			} else if err != nil {
+			if IsTillerNotFound(err) || IsTooManyResults(err) || IsTillerInvalidVersion(err) {
 				return microerror.Mask(err)
+			} else if err != nil {
+				return backoff.Permanent(microerror.Maskf(tillerStartingError, "can't establish a tunnel through new tiller pod with error %#q", err))
```
rossf7:
```suggestion
				return backoff.Permanent(microerror.Maskf(tillerStartingError, "can't establish  tunnel to tiller pod with error %#q", err))
```

So you're going to match on this error in the tiller resources?

tomahawk28: @rossf7 Yes, I will use it on tiller resource and would cancel reconciliation.

ensure_tiller_installed.go
TheoBrigitte: If it can recovers after `tooManyResultsError`, then fine.
I agree with ross here this has to be well tested.

tomahawk28:
Tested on `gorgoth`, works fine.

PR #133, opened 2019-11-11
Update and improve README

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7584

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,24 @@
-[![CircleCI](https://circleci.com/gh/giantswarm/helmclient.svg?&style=shield)](https://circleci.com/gh/giantswarm/helmclient)
+[![GoDoc](https://godoc.org/github.com/giantswarm/helmclient?status.svg)](http://godoc.org/github.com/giantswarm/helmclient) [![CircleCI](https://circleci.com/gh/giantswarm/helmclient.svg?&style=shield)](https://circleci.com/gh/giantswarm/helmclient)
 
 # helmclient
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,27 +1,24 @@
...
+[![GoDoc](https://godoc.org/github.com/giantswarm/helmclient?status.svg)](http://godoc.org/github.com/giantswarm/helmclient) [![CircleCI](https://circleci.com/gh/giantswarm/helmclient.svg?&style=shield)](https://circleci.com/gh/giantswarm/helmclient)
 
 # helmclient
 
-Package helmclient implements Helm related primitives to work against Tiller.
+Package helmclient implements [Helm] related primitives to work against helm
+releases. Currently supports Helm 2 and connects to the Tiller gRPC API using
+a port forwarding connection.
 
-## Projects using `helmclient`
```
rossf7: This projects using is out of date and a pain to keep updated. I think it's best to just drop it.

README.md
```diff
@@ -1,27 +1,24 @@
...
+## Interface
 
-### Test libraries
+See `helmclient.Interface` in [spec.go] for supported methods.
```
rossf7: I think the godoc for the interface is the best way to get an overview of what the lib supports. So trying to point people in that direction.

rossf7:
> I like the godoc badge. We should have it in all go projects.

Same here. I noticed Pawel added it in https://github.com/giantswarm/template/pull/3. I'm planning on working through the libs to improve the readmes. So I can do this then.

PR #134, opened 2019-11-19
Vendor dependencies, update k8s to 1.16.3

sslavic:
Towards giantswarm/giantswarm#7675


Gopkg.toml
```diff
@@ -76,23 +76,23 @@ required = [
...
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.15.4"
+  version = "kubernetes-1.16.3"
 
 [[constraint]]
   name = "k8s.io/helm"
-  version = "2.14.3"
+  version = "2.16.1"
```
rossf7: Quick drive by but client and server are tightly coupled. So for e2e to pass you also need to change the tiller image in `spec.go`.

helmclient.go
```diff
@@ -272,8 +272,8 @@ func (c *Client) getReleaseHistory(ctx context.Context, releaseName string) (*Re
 		}
 	}
 
-	if len(resp.Releases) > 1 {
-		return nil, microerror.Maskf(tooManyResultsError, "%d releases found, expected 1", len(resp.Releases))
+	if len(resp.Releases) == 0 {
```
tomahawk28: 👍

sslavic: Too many results cannot happen any more with fake client, it respects max history limit https://github.com/helm/helm/commit/ff3dc9dd33f3a31b6bbc3681d2853c03cd57ebdf#diff-ff36f764520e39db4e97a207158b51b2

I assume it was never possible with real client/tiller, that they respect the limit set in release history request, that we had it just because of bug in FakeClient 2.14 and older.

rossf7: I don't remember the details but that sounds totally reasonable. 👍

kopiczko:
While you are here can you also add:

```
[[constraint]]
  name = "github.com/Masterminds/semver"
  version = "=3.0.2"
```

I would create a PR but I don't want to introduce conflicts.

PR #135, opened 2019-11-20
Use permanent backoff for too many tillers error

rossf7:
Towards giantswarm/giantswarm#7603

When testing https://github.com/giantswarm/cluster-operator/pull/787 found a problem with the fix.

Due to the backoff the TooManyResults error is not returned. Getting it straight.

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -352,7 +352,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 				installTiller = true
 				return nil
 			} else if IsTooManyResults(err) {
-				return backoff.Permanent(err)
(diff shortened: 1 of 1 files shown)


ensure_tiller_installed.go
```diff
@@ -351,6 +351,8 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 				// Fall through as we need to install Tiller.
 				installTiller = true
 				return nil
+			} else if IsTooManyResults(err) {
+				return backoff.Permanent(err)
```
xh3b4sd: Can we mask the error given to `backoff.Permanent`? Then we do not miss out on the location in the stack.

rossf7: Thx good catch! Done.

PR #138, opened 2019-11-28
Make tiller image registry configurable

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7773

In AWS China after the pause container the next image we hit on cluster creation is Tiller. So making it configurable for use in app-operator.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -54,7 +54,8 @@ type Config struct {
 
 	EnsureTillerInstalledMaxWait time.Duration
 	RestConfig                   *rest.Config
-	TillerImage                  string
+	TillerImageName              string
+	TillerImageRegistry          string
 	TillerNamespace              string
 }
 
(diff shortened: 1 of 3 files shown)


helmclient.go
```diff
@@ -54,7 +54,8 @@ type Config struct {
 
 	EnsureTillerInstalledMaxWait time.Duration
 	RestConfig                   *rest.Config
-	TillerImage                  string
+	TillerImageName              string
+	TillerImageRegistry          string
```
rossf7: Splitting the tiller image so the registry can be overridden in China.

helmclient.go
```diff
@@ -102,6 +106,9 @@ func New(config Config) (*Client, error) {
 		Timeout: time.Second * httpClientTimeout,
 	}
 
+	// Registry is configurable for AWS China.
+	tillerImage := fmt.Sprintf("%s/%s", config.TillerImageRegistry, config.TillerImageName)
```
rossf7: We only used the TillerImage setting in the unit tests so it's safe to change.

helmclient_test.go
```diff
@@ -569,7 +570,7 @@ func Test_isTillerInvalidVersion(t *testing.T) {
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{
 						{
-							Image: defaultTillerImage,
+							Image: fmt.Sprintf("%s/%s", defaultTillerImageRegistry, defaultTillerImageName),
```
rossf7: Just making the tests work. Helm 3 is coming 🙏

rossf7:
Tested on giraffe.

PR #139, opened 2019-12-02
Migrate PSP API group use from deprecated extensions to policy

sslavic:
This PR minimizes https://github.com/giantswarm/helmclient/pull/134 with stuff that can be shipped independently from 1.16 upgrade, like the migration of long time deprecated API groups which are being dropped in 1.16.

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -140,7 +140,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			Rules: []rbacv1.PolicyRule{
 				{
 					APIGroups: []string{
-						"extensions",
+						"policy",
 					},
 					Resources: []string{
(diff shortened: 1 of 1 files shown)


ensure_tiller_installed.go
```diff
@@ -140,7 +140,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			Rules: []rbacv1.PolicyRule{
 				{
 					APIGroups: []string{
-						"extensions",
+						"policy",
```
tomahawk28: What if TC/CPs are still based on 1.14.x version k8s? is that still okay?

sslavic: Yes, because policy/v1beta1 is available since 1.10 (see https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)

PR #142, opened 2019-12-10
Check whether upgrade tiller or not

tomahawk28:
Towards giantswarm/giantswarm#7675

`app-operator` only need to create initial tiller pod, but should not upgrade it since we want to avoid race condition against tiller version.

### DIFF
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -396,7 +396,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 		}
 	} else if !installTiller && upgradeTiller {
 		if !upgrading {
-			c.logger.LogCtx(ctx, "level", "debug", "message", "found out-dated tiller but opt to not upgrading it.")
+			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but opt not to upgrade it.")
 			return nil
 		}
 		err = c.upgradeTiller(ctx, i)
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -22,8 +22,8 @@ import (
 // for it to start. A service account and cluster role binding are also created.
 // As a first step, it checks if Tiller is already ready, in which case it
 // returns early.
-func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
-	return c.EnsureTillerInstalledWithValues(ctx, []string{})
+func (c *Client) EnsureTillerInstalled(ctx context.Context, upgrading bool) error {
+	return c.EnsureTillerInstalledWithValues(ctx, []string{}, upgrading)
 }
 
 // EnsureTillerInstalledWithValues installs Tiller by creating its deployment
@@ -31,7 +31,7 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 // also created. As a first step, it checks if Tiller is already ready, in
 // which case it returns early. Values can be provided to pass through to Tiller
 // and overwrite its deployment.
-func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []string) error {
+func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []string, upgrading bool) error {
 	// Check if Tiller is already present and return early if so.
 	{
 		c.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding if tiller is installed in namespace %#q", c.tillerNamespace))
@@ -395,6 +395,9 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			return microerror.Mask(err)
 		}
 	} else if !installTiller && upgradeTiller {
+		if !upgrading {
+			return nil
+		}
 		err = c.upgradeTiller(ctx, i)
 		if err != nil {
 			return microerror.Mask(err)
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -395,6 +395,9 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			return microerror.Mask(err)
 		}
 	} else if !installTiller && upgradeTiller {
+		if !c.upgrading {
+			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but opt not to upgrade it.")
+		}
 		err = c.upgradeTiller(ctx, i)
 		if err != nil {
 			return microerror.Mask(err)
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -22,8 +22,8 @@ import (
 // for it to start. A service account and cluster role binding are also created.
 // As a first step, it checks if Tiller is already ready, in which case it
 // returns early.
-func (c *Client) EnsureTillerInstalled(ctx context.Context, upgrading bool) error {
-	return c.EnsureTillerInstalledWithValues(ctx, []string{}, upgrading)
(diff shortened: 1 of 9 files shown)


ensure_tiller_installed.go
```diff
@@ -395,6 +395,10 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			return microerror.Mask(err)
 		}
 	} else if !installTiller && upgradeTiller {
+		if !upgrading {
+			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but opt not to upgrade it.")
```
tomahawk28: We can return earlier than this actually, but I just want to place much simpler logic in here.

spec.go
```diff
@@ -38,12 +38,12 @@ type Interface interface {
 	// for it to start. A service account and cluster role binding are also created.
 	// As a first step, it checks if Tiller is already ready, in which case it
 	// returns early.
-	EnsureTillerInstalled(ctx context.Context) error
+	EnsureTillerInstalled(ctx context.Context, upgrading bool) error
```
tomahawk28: I would like to avoid naming dispute since these would eventually be deleted on helm v3.

ensure_tiller_installed.go
```diff
@@ -22,16 +22,16 @@ import (
 // for it to start. A service account and cluster role binding are also created.
 // As a first step, it checks if Tiller is already ready, in which case it
 // returns early.
-func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
-	return c.EnsureTillerInstalledWithValues(ctx, []string{})
+func (c *Client) EnsureTillerInstalled(ctx context.Context, upgrading bool) error {
```
rossf7: Won't this change the signature? 

So we'd need to change it everywhere we call `EnsureTillerInstalled`? This is used in lots of e2e tests. What about adding it as an optional parameter? 

https://github.com/giantswarm/helmclient/blob/d3c47349776d30db46c3e200ee94d16aeb22aa0b/helmclient.go#L55-L59

We could default it to false and only enable for draughtsman and chart-operator where we want the upgrade to happen.

tomahawk28: that's nice, I'm moving it from a parameter to config. PTAL!

ensure_tiller_installed.go
```diff
@@ -395,6 +395,9 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			return microerror.Mask(err)
 		}
 	} else if !installTiller && upgradeTiller {
+		if !c.upgrading {
+			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but opt not to upgrade it.")
```
rossf7:
```suggestion
			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but upgrades not enabled")
```

helmclient.go
```diff
@@ -57,6 +57,7 @@ type Config struct {
 	TillerImageName              string
 	TillerImageRegistry          string
 	TillerNamespace              string
+	Upgrading                    bool
```
rossf7:
```suggestion
	TillerUpgradeEnabled                    bool
```

Can we rename it to this? Otherwise looks good.

ensure_tiller_installed.go
```diff
@@ -395,6 +395,9 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 			return microerror.Mask(err)
 		}
 	} else if !installTiller && upgradeTiller {
+		if !c.tillerUpgradeEnabled {
+			c.logger.LogCtx(ctx, "level", "debug", "message", "found an out-dated tiller but upgrades not enabled")
```
rossf7: How does this return early? Won't upgradeTiller still get called?

tomahawk28: 🤦‍♂ Totally forgot it, Thanks to point out.

tomahawk28:
By failing e2e test, we actually test it 😆 
{"caller":"github.com/giantswarm/helmclient/ensure_tiller_installed.go:399","level":"debug","message":"found an out-dated tiller but upgrades not enabled","time":"2019-12-10T16:26:22.400251+00:00"}

tomahawk28:
Found some issues after vendoring on app-operator 
```
W 12/11 09:59:37 /apis/application.giantswarm.io/v1alpha1/namespaces/imm44/apps/chart-operator chartoperatorv1 retrying backoff in '5s' due to error | backoff/notifier.go:13 | event=update | version=114033467
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:205:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:687:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:791: current tiller version `2.14.3` does not match desired tiller version `2.16.1`
	tiller invalid version error
W 12/11 09:59:42 /apis/application.giantswarm.io/v1alpha1/namespaces/imm44/apps/chart-operator chartoperatorv1 retrying backoff in '5s' due to error | backoff/notifier.go:13 | event=update | version=114033467
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:205:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:687:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:791: current tiller version `2.14.3` does not match desired tiller version `2.16.1`
	tiller invalid version error
W 12/11 09:59:47 /apis/application.giantswarm.io/v1alpha1/namespaces/imm44/apps/chart-operator chartoperatorv1 retrying backoff in '5s' due to error | backoff/notifier.go:13 | event=update | version=114033467
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:205:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:687:
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:791: current tiller version `2.14.3` does not match desired tiller version `2.16.1`
	tiller invalid version error
```

rossf7:
@tomahawk28 Some errors are expected due to the retry logic. Is there a failing build you can point me at?

tomahawk28:
No failed builds but just saw this on `ghost` after deployment of https://github.com/giantswarm/app-operator/pull/175
But I think this is not okay since some TCs would remain at lower version in tiller, but app-operator would expect the higher version anyway.

tomahawk28:
@rossf7 I will try to ignore out-dated tiller when we create a tunnel. WDYT?

rossf7:
Let's handle the tiller version error and cancel the resource. The upgrade process is 

- We deploy the new chart-operator in the tenant cluster.
- chart-operator upgrades tiller.
- We no longer get the errors in app-operator.

If the tiller upgrade fails then the tiller unreacheable alert will fire. So it's safe to just handle the error and cancel the resource.

rossf7:
> I will try to ignore out-dated tiller when we create a tunnel.

@tomahawk28 Argh page didn't update :( If that works then fine but I'm also good with returning an error and matching for it in app-operator.

tomahawk28:
@rossf7 The problem is, app-operator have only `v1` version for each controller, and the latest app-oprator would handle tillers in old TCs, too. So we can't just cancel the resource or reconciliation.

tomahawk28:
Tested on `ghost`

PR #143, opened 2019-12-18
Return separate error for 404s pulling tarballs

rossf7:
Sorting out the error handling.

### DIFF
--- a/error.go
+++ b/error.go
@@ -95,6 +95,15 @@ func IsPullChartFailedError(err error) bool {
 	return microerror.Cause(err) == pullChartFailedError
 }
 
+var pullChartNotFoundError = &microerror.Error{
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -95,6 +95,15 @@ func IsPullChartFailedError(err error) bool {
 	return microerror.Cause(err) == pullChartFailedError
 }
 
+var pullChartNotFoundError = &microerror.Error{
+	Kind: "pullChartNotFoundError",
+}
```
rossf7: Currently we are returning the same error for host not found and 404s for missing tarballs.

I think this is wrong because host not found is due to an incorrect appcatalog CR. But we can get tarball 404s due to timing issues.

PR #146, opened 2019-12-30
Use configured backoff time in EnsureTillerInstalled

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8190

We will configure the max wait to 30 secs instead of 2 mins in app-operator. But I couple of the retries missed this setting.


rossf7:
> why was this hardcoded before though

The setting was added a lot later. 😄 Anyway all this shite goes away with Helm 3.

PR #147, opened 2020-01-13
Handle additional errors

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7355

Handles additional error cases. Used in https://github.com/giantswarm/chart-operator/pull/343.

- Chart tarball has an empty templates directory.
- Helm release name is invalid.
- Helm release is not deployed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ e2eTest: &e2eTest
     - run:
         name: Get helm binary
         command: |
-          curl -L https://get.helm.sh/helm-v2.14.3-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
+          curl -L https://get.helm.sh/helm-v2.16.0-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
           chmod +x helm
 
     - run:
@@ -34,7 +34,7 @@ e2eTest: &e2eTest
           then
             export KUBECONFIG=$(./e2ectl kubeconfig path)
             ./kubectl create ns giantswarm
-            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.14.3 --wait
+            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.16.0 --wait
           fi
 
     - run:
--- a/error.go
+++ b/error.go
@@ -33,6 +33,32 @@ func IsCannotReuseRelease(err error) bool {
 	return false
 }
 
+var (
+	emptyChartTemplatesRegexp = regexp.MustCompile(`release \S+ failed: no objects visited`)
+)
+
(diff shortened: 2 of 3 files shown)


error.go
```diff
@@ -33,6 +33,32 @@ func IsCannotReuseRelease(err error) bool {
 	return false
 }
 
+var (
+	emptyChartTemplatesRegexp = regexp.MustCompile(`release \S+ failed: no objects visited`)
```
rossf7: `no objects visited` is pretty cryptic. So I've named the error `emptyChartTemplates` as this is the only time we see this so far.

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ e2eTest: &e2eTest
     - run:
         name: Get helm binary
         command: |
-          curl -L https://get.helm.sh/helm-v2.14.3-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
+          curl -L https://get.helm.sh/helm-v2.16.0-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
```
sslavic:
```suggestion
          curl -L https://get.helm.sh/helm-v2.16.1-linux-amd64.tar.gz | tar xvz --strip-components 1 linux-amd64/helm
```

.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ e2eTest: &e2eTest
           then
             export KUBECONFIG=$(./e2ectl kubeconfig path)
             ./kubectl create ns giantswarm
-            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.14.3 --wait
+            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.16.0 --wait
```
sslavic:
```suggestion
            ./helm init --tiller-namespace giantswarm --tiller-image quay.io/giantswarm/tiller:v2.16.1 --wait
```

.circleci/config.yml
rossf7: @sslavic This is for the tiller upgrade test. I needed to bump the version because 2.14.3 won't install in KIND as its K8s 1.16 now.

sslavic: Ah, clear, thanks for heads up. 2.16.0 had some bug, so they quickly came out with 2.16.1, hopefully it won't affect these tests.

rossf7: We call `EnsureTillerInstalled` in the tests so tiller is upgraded to 2.16.1 before the tests run. 

e.g. https://github.com/giantswarm/helmclient/blob/8d2f16a18b92557fea612dcae25da5c317739b51/integration/test/basic/basic_test.go#L26

We should maybe only install tiller manually for the upgrade tiller test. But I don't want to invest too much time here. As this goes away with the upgrade to Helm 3.

PR #154, opened 2020-01-22
Remove Helm 2 code

rossf7:
Towards giantswarm/roadmap#30

This PR starts to remove the Helm 2 specific code. The changes are quite big so I'll split this into a few PRs.

- `ensure_tiller_installed.go` is removed entirely.
- `helmclient_test.go` is also removed entirely.

The unit tests are complex and make heavy use of the fake client. Coverage will be replaced with integration tests.


helmclient.go
```diff
@@ -146,32 +136,7 @@ func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options
...
-	}
-	b := backoff.NewMaxRetries(10, 5*time.Second)
-	n := backoff.NewNotifier(c.logger, ctx)
-
-	err := backoff.RetryNotify(o, b, n)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
+	c.logger.LogCtx(ctx, "level", "debug", "message", "delete release not yet implemented for helm 3")
```
rossf7: Helm 2 implementation logic is removed. It will be replaced with the new Helm 3 logic.

PR #155, opened 2020-01-27
Handle timeouts pulling charts and add monitoring

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8573

Adds missing monitoring to PullChartTarball and handles timeout errors which we see sometimes in multiple installations.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,6 +1,7 @@
 package helmclient
 
 import (
+	"net"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -130,6 +131,34 @@ func IsPullChartNotFound(err error) bool {
...
+	if c == pullChartTimeoutError {
+		return true
+	}
+
+	netErr, ok := err.(net.Error)
+	if !ok {
+		return false
+	}
+
+	if netErr.Timeout() {
```
tomahawk28: What about `return netErr.Timeout()`? since it's bool returning anyway.

pull_chart_tarball.go
```diff
@@ -11,12 +11,28 @@ import (
...
 // returning the file path.
 func (c *Client) PullChartTarball(ctx context.Context, tarballURL string) (string, error) {
+	eventName := "pull_chart_tarball"
+
+	t := prometheus.NewTimer(histogram.WithLabelValues(eventName))
+	defer t.ObserveDuration()
+
+	chartTarballPath, err := c.pullChartTarball(ctx, tarballURL)
+	if err != nil {
+		errorGauge.WithLabelValues(eventName).Inc()
```
tomahawk28: Can we report chart names or at least server name (`quay.io`) as well?

rossf7: I can see the chart names or even the tarball URL as being useful. As we could alert on charts with 404s.

But all the metrics are the same and just take the event name. I don't have more time to spend on this. But maybe we can improve it as part of the Helm 3 changes?

error.go
rossf7: Nice. Done.

pull_chart_tarball.go
tomahawk28: Sure, That would be cool.

rossf7: Thanks. I added it to the checklist in the Helm 3 ticket.

PR #156, opened 2020-01-28
Remove rest of Helm 2 code and align interface with Helm 3

rossf7:

This is a bit big but hopefully understandable. The interface changes are easier to see in `spec.go`. 

Once all the Helm 2 code is gone I think it will be easier to migrate to modules and then add Helm 3.


helmclient.go
```diff
@@ -76,13 +73,13 @@ func New(config Config) (*Client, error) {
 }
 
 // DeleteRelease uninstalls a chart given its release name.
-func (c *Client) DeleteRelease(ctx context.Context, releaseName string, options ...helmclient.DeleteOption) error {
+func (c *Client) DeleteRelease(ctx context.Context, releaseName string) error {
```
rossf7: For Helm 2 we set `helm.DeletePurge(true)` but this option was removed from Helm 3.

https://github.com/giantswarm/chart-operator/blob/86901f94d21f395f540102afa91be92ac7607ee0/service/controller/chart/v1/resource/release/delete.go#L21

helmclient.go
```diff
@@ -143,13 +140,13 @@ func (c *Client) getReleaseHistory(ctx context.Context, releaseName string) (*Re
 }
 
 // InstallReleaseFromTarball installs a chart packaged in the given tarball.
-func (c *Client) InstallReleaseFromTarball(ctx context.Context, path, ns string, options ...helmclient.InstallOption) error {
+func (c *Client) InstallReleaseFromTarball(ctx context.Context, chartPath string, values map[string]interface{}, options InstallOptions) error {
```
rossf7: The values used to be an option. Now they are passed as a `map[string]interface{}`. `helmclient.MergeValues` already returns this so we are good.

https://github.com/giantswarm/helmclient/blob/31b69ea684193307663277dd36a69ed0e0ef3822/merge_values.go#L22

helmclient.go
```diff
@@ -232,20 +220,20 @@ func (c *Client) RunReleaseTest(ctx context.Context, releaseName string, options
...
-func (c *Client) runReleaseTest(ctx context.Context, releaseName string, options ...helmclient.ReleaseTestOption) error {
+func (c *Client) runReleaseTest(ctx context.Context, releaseName string, options ReleaseTestOptions) error {
 	c.logger.LogCtx(ctx, "level", "debug", "message", "run release test not yet implemented for helm 3")
 	return nil
 }
 
 // UpdateReleaseFromTarball updates the given release using the chart packaged
 // in the tarball.
-func (c *Client) UpdateReleaseFromTarball(ctx context.Context, releaseName, path string, options ...helmclient.UpdateOption) error {
+func (c *Client) UpdateReleaseFromTarball(ctx context.Context, releaseName, chartPath string, values map[string]interface{}, options UpdateOptions) error {
```
rossf7: Same interface change for update.

spec.go
```diff
@@ -39,8 +37,29 @@ type Interface interface {
...
 	// RunReleaseTest runs the tests for a Helm Release. This is the same
 	// action as running the helm test command.
-	RunReleaseTest(ctx context.Context, releaseName string, options ...helm.ReleaseTestOption) error
+	RunReleaseTest(ctx context.Context, releaseName string, options ReleaseTestOptions) error
 	// UpdateReleaseFromTarball updates the given release using the chart packaged
 	// in the tarball.
-	UpdateReleaseFromTarball(ctx context.Context, releaseName, path string, options ...helm.UpdateOption) error
+	UpdateReleaseFromTarball(ctx context.Context, releaseName, chartPath string, values map[string]interface{}, options UpdateOptions) error
+}
+
```
rossf7: These are the options we're using currently. I don't think we need to support all the options.

I think its cleaner to define the options ourselves rather than using the helm types as we did previously.

PR #157, opened 2020-01-29
helm3: Handle timeouts pulling charts and add monitoring

rossf7:
Cherry pick of https://github.com/giantswarm/helmclient/pull/155.

PR #158, opened 2020-01-29
Use context.WithTimeout to limit time connecting to tiller

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8479

This looks fine tested locally but I want to test tomorrow with my nemesis `asgard/f24sd`. To make sure it's fixed this time.

```
D 01/29 17:22:58 /apis/application.giantswarm.io/v1alpha1/namespaces/f24sd/apps/chart-operator tillerv1 finding if tiller is installed in namespace `giantswarm` | helmclient/ensure_tiller_installed.go:38 | controller=app-operator | event=update | loop=52077 | version=186624200
D 01/29 17:42:58 /apis/application.giantswarm.io/v1alpha1/namespaces/f24sd/apps/chart-operator tillerv1 found that tiller is not installed in namespace `giantswarm` | helmclient/ensure_tiller_installed.go:44 | controller=app-operator | event=update | loop=52077 | version=186624200
```

app-operator tried to check if tiller is installed but it gets stuck for 20 mins. :( This uses context.WithTimeout to ensure we only wait for 10 seconds.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -54,9 +54,9 @@ required = [
   name = "github.com/giantswarm/errors"
   branch = "master"
 
-[[constraint]]
+[[override]]
   name = "github.com/giantswarm/k8sportforward"
-  branch = "master"
+  branch = "tiller-again"
 
 [[constraint]]
   name = "github.com/giantswarm/microerror"
--- a/ensure_tiller_installed.go
+++ b/ensure_tiller_installed.go
@@ -41,7 +41,7 @@ func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []s
 
 		c.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding if tiller is installed in namespace %#q", c.tillerNamespace))
 
-		t, err := c.newTunnel()
+		t, err := c.newTunnel(ctx)
 		defer c.closeTunnel(ctx, t)
 		if err != nil {
 			// fall through, we may need to create or upgrade Tiller.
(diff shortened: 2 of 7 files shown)


ensure_tiller_installed.go
```diff
@@ -13,7 +14,7 @@ import (
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
rossf7: Fixing the import.

ensure_tiller_installed.go
```diff
@@ -35,6 +36,9 @@ func (c *Client) EnsureTillerInstalled(ctx context.Context) error {
 func (c *Client) EnsureTillerInstalledWithValues(ctx context.Context, values []string) error {
 	// Check if Tiller is already present and return early if so.
 	{
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
```
rossf7: We do this in the migration resources in cluster-operator and it works well. Giving it 10 secs because of the dumb port forwarding connection.

Gopkg.toml
```diff
@@ -54,9 +54,9 @@ required = [
   name = "github.com/giantswarm/errors"
   branch = "master"
 
-[[constraint]]
+[[override]]
   name = "github.com/giantswarm/k8sportforward"
-  branch = "master"
+  branch = "tiller-again"
```
MarcelMue: This needs to be reverted before merging, right?

rossf7: Yes but these changes still don't work sadly. Sorry I asked for review too early. It worked in ginger.

rossf7:
Doesn't work. Sorry about the noise. Closing in favor of #157 which I still want to test in asgard to be sure its fixed.

PR #159, opened 2020-01-30
Don't try to create tunnel when tiller is outdated

rossf7:
Towards giantswarm/giantswarm#8479

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -202,6 +202,8 @@ func (c *Client) getReleaseContent(ctx context.Context, releaseName string) (*Re
 			t, err := c.newTunnel()
 			if IsTillerNotFound(err) {
 				return backoff.Permanent(microerror.Mask(err))
+			} else if IsTillerOutdated(err) {
+				return backoff.Permanent(microerror.Mask(err))
 			} else if IsEmptyChartTemplates(err) {
 				return backoff.Permanent(microerror.Mask(err))
 			} else if IsReleaseNameInvalid(err) {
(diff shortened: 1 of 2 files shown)


helmclient.go
```diff
@@ -698,7 +700,7 @@ func (c *Client) newTunnel() (*k8sportforward.Tunnel, error) {
 	err = validateTillerVersion(pod, c.tillerImage)
 	if err != nil {
 		if IsTillerInvalidVersion(err) && !c.tillerUpgradeEnabled {
-			c.logger.Log("level", "debug", "message", "found an out-dated tiller but keep going to create a tunnel")
+			return nil, microerror.Maskf(tillerOutdatedError, "found an out-dated tiller but upgrades not enabled")
```
rossf7: Tiller is upgraded by chart-operator in the tenant cluster. When we want to upgrade Tiller we deploy a new version of chart-operator.

So when upgrades are disabled in app-operator we should return the error and cancel the resource.

rossf7:
Tested on `ginger` and with `asgard/f24sd`.

rossf7:
Going with this to get app-operator problems in asgard sorted.

xh3b4sd:
I see we fixed a lot of bugs in the past weeks here. Is there a light at the end of the tunnel or will this go on for some time?

rossf7:
The current helmclient has many issues due to Tiller and the crappy port forwarding connection. So I'm just fixing the high priority bugs.

This is why the Helm 3 upgrade involves a rewrite of helmclient. I hope we can learn from some of the mistakes we made the first time.

The spec is here. I'd love another couple of approvals before merging.

https://github.com/giantswarm/giantswarm/pull/8458

PR #164, opened 2020-02-13
helm3: Implement GetReleaseContent

rossf7:

Implements the first helm 3 method.


integration/test/basic/basic_test.go
```diff
@@ -4,78 +4,20 @@ package basic
...
-	if err != nil {
-		t.Fatalf("could not delete release %v", err)
-	}
-
-	releaseContent, err = config.HelmClient.GetReleaseContent(ctx, releaseName)
-	if err != nil {
-		t.Fatalf("could not get release content %v", err)
+	releaseContent, err := config.HelmClient.GetReleaseContent(ctx, "no-release-exists", "kube-system")
+	if err != nil && !helmclient.IsReleaseNotFound(err) {
+		t.Fatalf("expected release not found error got %v", err)
```
rossf7: Simplified the integration test for now. It will be extended as more logic is added.

get.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/release"
+)
+
+// GetReleaseContent gets the current status of the Helm Release including any
+// values provided when the chart was installed. The releaseName is the name
+// of the Helm Release that is set when the Helm Chart is installed.
+func (c *Client) GetReleaseContent(ctx context.Context, releaseName, namespace string) (*ReleaseContent, error) {
```
rossf7: Method signature needs to change to take in the namespace as well as the release name.

This is so the helm kube client is scoped to the namespace. Before we just configured the tiller namespace.

PR #165, opened 2020-02-17
helm3: Delete tiller errors and update readme

rossf7:
Some tidy up prior to #164


rossf7:
Not controversial. Going with a ping.

PR #166, opened 2020-02-17
helm3:  Implement InstallChartFromTarball

rossf7:

### DIFF
--- a/install.go
+++ b/install.go
@@ -1,0 +1,58 @@
+package helmclient
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart/loader"
+)
+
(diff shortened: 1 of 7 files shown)


install.go
```diff
@@ -0,0 +1,58 @@
...
+	err := c.installReleaseFromTarball(ctx, chartPath, namespace, values, options)
+	if err != nil {
+		errorGauge.WithLabelValues(eventName).Inc()
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (c *Client) installReleaseFromTarball(ctx context.Context, chartPath, namespace string, values map[string]interface{}, options InstallOptions) error {
```
xh3b4sd: For my curiosity, why are the values not options in that sense?

rossf7: The Helm API changed. In Helm 2 there was a ValueOverrides option.

In Helm 3 the install action takes in a chart and values. So I'd like to align with that.

https://github.com/helm/helm/blob/v3.1.0/pkg/action/install.go#L163

PR #168, opened 2020-02-17
Fix param order for GetReleaseContent

rossf7:
I got the ordering wrong. 🤦‍♂


rossf7:
Going with a ping. Sorry for the noise.

PR #170, opened 2020-02-18
helm3: Implement LoadChart and ListReleases

rossf7:
Towards giantswarm/roadmap#30

### DIFF
--- a/integration/test/basic/basic_test.go
+++ b/integration/test/basic/basic_test.go
@@ -19,6 +19,9 @@ import (
 func TestBasic(t *testing.T) {
 	ctx := context.Background()
 
+	var err error
+	var releaseName string = "test-chart"
+
 	{
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "checking release not found")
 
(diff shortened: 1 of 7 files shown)


loadchart.go
```diff
@@ -0,0 +1,47 @@
...
+	return newChart(chartRequested)
+}
+
+func newChart(helmChart *chart.Chart) (Chart, error) {
+	if helmChart == nil || helmChart.Metadata == nil {
+		return Chart{}, microerror.Maskf(executionFailedError, "expected non nil argument but got %#v", helmChart)
+	}
+
+	chart := Chart{
+		Version: helmChart.Metadata.Version,
```
rossf7: We use LoadChart to calculate the desired state in chart-operator. So so far we only care about the chart version.

tomahawk28:
LGTM but there's a build error 
```
# github.com/giantswarm/helmclient/helmclienttest [github.com/giantswarm/helmclient/helmclienttest.test]
helmclienttest/helmclient.go:40:2: cannot use c (type *Client) as type helmclient.Interface in return argument:
	*Client does not implement helmclient.Interface (wrong type for ListReleaseContents method)
		have ListReleaseContents(context.Context) ([]*helmclient.ReleaseContent, error)
		want ListReleaseContents(context.Context, string) ([]*helmclient.ReleaseContent, error)
```

rossf7:
Oof I forgot to update the test package. Thanks! I’ll fix it after lunch.

integration/test/basic/basic_test.go
```diff
@@ -19,6 +19,9 @@ import (
 func TestBasic(t *testing.T) {
 	ctx := context.Background()
 
+	var err error
+	var releaseName string = "test-chart"
+
```
tomahawk28:
```suggestion
```

integration/test/basic/basic_test.go
```diff
@@ -33,29 +36,73 @@ func TestBasic(t *testing.T) {
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "checked release not found")
 	}
 
-	var releaseName string = "test-chart"
+	var chartPath string
```
tomahawk28:
```suggestion
```

rossf7:
@tomahawk28 Thanks for approving. 

But I disagree the whitespace needs to change. This is the same block style we use in e2etests.

https://github.com/giantswarm/e2etests/blob/208247dafcdeff1443690d271cb9e033c1e134e1/clusterstate/clusterstate.go

Going with this.

PR #177, opened 2020-03-11
Use integration-test job from architect-orb

rossf7:
Towards giantswarm/giantswarm#4606


rossf7:
Closing. Need to sort path errors with charttarball package.

PR #179, opened 2020-03-16
helm3:  Add constants for statuses

rossf7:
Towards giantswarm/roadmap#30

The Helm release statuses change with Helm 3. In chart-operator the strings are hardcoded in lots of places.

This adds them as constants here. The rationale is so we can have constants but without importing helm packages.

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -35,6 +35,17 @@ const (
 	StatusPendingRollback = "pending-rollback"
 )
 
+var (
+	// ReleaseTransitionStatuses is used to determine if the Helm Release is
+	// currently being updated.
+	ReleaseTransitionStatuses = map[string]bool{
+		StatusUninstalled:     true,
+		StatusPendingInstall:  true,
+		StatusPendingUpgrade:  true,
+		StatusPendingRollback: true,
+	}
+)
+
 const (
 	// defaultMaxHistory is the maximum number of release versions stored per
 	// release by default.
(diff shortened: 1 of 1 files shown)


spec.go
```diff
@@ -19,7 +56,7 @@ const (
 	runReleaseTestTimout = 300
 )
 
-// Interface describes the methods provided by the helm client.
+// Interface describes the methods provided by the Helm client.
```
rossf7: Some cleanup to improve the godoc.

spec.go
```diff
@@ -25,6 +25,10 @@ const (
 	StatusSuperseded = "superseded"
 	// StatusFailed indicates that the release was not successfully deployed.
 	StatusFailed = "failed"
+	// StatusNotInstalled indicates that an error occurred and Helm could not
+	// create a release. NOTE: This status is Giant Swarm specific and DOES
+	// NOT map to a Helm status.
+	StatusNotInstalled = "not installed"
```
pipo02mix: Out of curiosity, what is different from `failed` status?

rossf7: Yes I wasn't sure about this. So I went with keeping it as a constant in chart-operator.

Failed is helm created the release but there is a problem with it. Not installed is helm couldn't create a release. 

So it's not a true helm status which is why I left it out of helmclient. Let me know if you think it should be in helmclient.

PR #180, opened 2020-03-17
helm3: Get all values in GetReleaseContent

rossf7:
Towards giantswarm/roadmap#30

The Helm 2 API was weird because we needed to call `GetReleaseHistory` to get version info. 

This meant the status resource in chart-operator needed to call both `GetReleaseContent` and `GetReleaseHistory`. With Helm 3 we can get all the info in a single API call.

I'll do a follow up to extend `GetReleaseHistory` to use the new API but this is all we need for chart-operator.

Note: This also bumps Helm from `3.1.1` to `3.1.2` which came out last week.


rossf7:
Last helmclient PR before using it in the helm3 branch of chart-operator.

PR #198, opened 2020-04-07
Increase HTTP client timeout from 5 to 30 secs for testing

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9946

We currently use a hardcoded HTTP client timeout of 5 secs when pulling chart tarballs. This is regularly timing out in AWS China due to the slow network. 

This makes it configurable so it can be increased in these installations.

## Checklist


rossf7:
Tested on giraffe.

PR #215, opened 2020-05-18
Applying helm3 branch

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/10973

Applying `helm3` branches into master. Added Changelog entry. 

## Checklist


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed to support Helm 3; To keep using Helm 2, please use the version under 0.2.2.
```
tomahawk28, line 12: Is this should be added or changed? Since this is break changes, I prefer to have it as changed.

README.md
```diff
@@ -3,22 +3,16 @@
...
-a port forwarding connection.
+releases. Currently supports Helm 3.
 
 ## Interface
 
 See `helmclient.Interface` in [spec.go] for supported methods.
 
-## Related libraries
-
-[k8sportforward] is used to establish a port forwarding conection with Tiller.
```
tomahawk28, line 15: k8sportforward was deleted.

CHANGELOG.md
rossf7:
```suggestion
- Updated to support Helm 3; To keep using Helm 2, please use version 0.2.X.
```

Good point. What about this?  If we have to make fixes for Helm 2 the version will be 0.2.X.

tomahawk28: Thanks, that makes sense.

PR #229, opened 2020-06-09
Extending map merging errors

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11469

I will use new errors in chart-operator, to let users know they have a problem in their configmap/secret.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add new yaml parsing errors.
+
 ## [1.0.2] 2020-05-29
 
 - Add new validation error for both install and upgrade cases.
--- a/merge_values_test.go
+++ b/merge_values_test.go
@@ -62,6 +62,12 @@ nested: null`
 const nullValuedYaml = `
 null`
 
+const wrongYaml = `
+nested:
+    values: "nested"
+    - data: 
+"`
+
 func Test_MergeValues(t *testing.T) {
 	testCases := []struct {
 		name           string
@@ -142,7 +148,7 @@ func Test_MergeValues(t *testing.T) {
 				"another": []byte("test: val"),
 				"test":    []byte("test: val"),
 			},
-			errorMatcher: IsExecutionFailed,
+			errorMatcher: IsParsingSrcFailedError,
 		},
 		{
 			name: "case 6: dest map with multiple keys returns error",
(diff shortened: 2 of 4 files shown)


merge_values_test.go
```diff
@@ -142,7 +148,7 @@ func Test_MergeValues(t *testing.T) {
 				"another": []byte("test: val"),
 				"test":    []byte("test: val"),
 			},
-			errorMatcher: IsExecutionFailed,
+			errorMatcher: IsParsingSrcFailedError,
```
tomahawk28, line 151: If it's two key cm, I think we should indicate this is wrong to customers, by using app CRs. So this is fine.

merge_values_test.go
```diff
@@ -178,6 +184,16 @@ func Test_MergeValues(t *testing.T) {
...
 		},
+		{
+			name: "case 10: wrong dest yaml returns error",
+			destMap: map[string][]byte{
+				"test": []byte(wrongYaml),
+			},
+			srcMap: map[string][]byte{
+				"test": []byte("test: val"),
+			},
+			errorMatcher: IsParsingDestFailedError,
```
sslavic, line 195: What's going to be error message we'll actually see in this case?

tomahawk28: It will be like this, ```{"kind":"parsingDestFailedError","annotation":"yaml: line 3: did not find expected key","stack":[{"file":"/Users/jihyukbok/go/src/github.com/giantswarm/helmclient/merge_values.go","line":27}```
If we could ensure that whether the destination or source file is a problem, then we can show this message to users by pointing the actual problem file.

sslavic: Higher level component making use of helm client (app/chart-operator?) could add context (what was destination, and what was source), to make error message more meaningful.

tomahawk28: Sure, I'm testing helmclient and it results like this. 
```
...
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'parsing error: failed to parse user configmap, logs from merging: parsing
      src failed error: yaml: line 22: did not find expected key'
    status: configmap-merge-failed
  version: ""
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
sslavic, line 8: @tomahawk28 changelog needs some more updates with 1.0.3 release

PR #231, opened 2020-06-11
Apply Helm 2.16.8 security patch

tomahawk28:
## Checklist

- [ ] Update changelog in CHANGELOG.md.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.8.4
+  architect: giantswarm/architect@0.9.0
 
 workflows:
   test:
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/giantswarm/helmclient
 
-go 1.13
+go 1.14
 
 require (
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.8.4
+  architect: giantswarm/architect@0.9.0
```
tomahawk28, line 4: Only after architect-orb 0.8.15, we are upgrading go version to 1.14. So better to update this at the latest.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/giantswarm/helmclient
 
-go 1.13
+go 1.14
```
tomahawk28, line 3: Here we need to update go mod version, too.

spec.go
```diff
@@ -20,7 +20,7 @@ const (
 	// runReleaseTestTimeout is the timeout in seconds when running tests.
 	runReleaseTestTimout = 300
 
-	defaultTillerImageName     = "giantswarm/tiller:v2.16.1"
+	defaultTillerImageName     = "giantswarm/tiller:v2.16.8"
```
tomahawk28, line 23: Upgrade the base spec as 2.16.8

PR #237, opened 2020-06-22
Exposing Release namespace 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11651

There's no way to extract release namespace information from any other resources. 
Hence I'm back to change the spec of helmclient for 0.x.x.

## Checklist

- [ ] Update changelog in CHANGELOG.md.

### DIFF
--- a/helmclient.go
+++ b/helmclient.go
@@ -323,6 +323,7 @@ func (c *Client) getReleaseHistory(ctx context.Context, releaseName string) (*Re
 			Description:  release.Info.Description,
 			LastDeployed: lastDeployed,
 			Name:         release.Name,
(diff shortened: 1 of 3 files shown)


helmclient.go
```diff
@@ -323,6 +323,7 @@ func (c *Client) getReleaseHistory(ctx context.Context, releaseName string) (*Re
 			Description:  release.Info.Description,
 			LastDeployed: lastDeployed,
 			Name:         release.Name,
+			Namespace:    release.Namespace,
```
tomahawk28, line 326: Here we assigned the namespace value for helmclient.

PR #252, opened 2020-07-16
Add timeouts for installing or updating helm releases

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11735

Needed for https://github.com/giantswarm/opsctl/pull/740.

## Checklist

### DIFF
--- a/update.go
+++ b/update.go
@@ -2,7 +2,6 @@ package helmclient
 
 import (
 	"context"
-	"time"
 
 	"github.com/giantswarm/microerror"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


update.go
```diff
@@ -53,9 +53,14 @@ func (c *Client) updateReleaseFromTarball(ctx context.Context, chartPath, namesp
 }
 
 func (options UpdateOptions) configure(action *action.Upgrade, namespace string) {
+	//if options.Timeout == 0 {
```
tomahawk28, line 56: We can delete those comments?

rossf7: Good catch. I think we should keep them to be sure there is a timeout in place.

PR #262, opened 2020-08-07
Disable OpenAPI validation

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11748

Getting this in for when we create the new release. If we need this to be configurable I think we can add that later. WDYT?

## Checklist


rossf7:
Going with a ping.

PR #271, opened 2020-09-21
Add metrics for event counts

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/13309

## Checklist

### DIFF
--- a/pkg/helmclient/rollback.go
+++ b/pkg/helmclient/rollback.go
@@ -22,6 +22,8 @@ func (c *Client) Rollback(ctx context.Context, namespace, releaseName string, re
 		return microerror.Mask(err)
 	}
 
+	eventCounter.WithLabelValues(eventName).Inc()
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/helmclient/rollback.go
```diff
@@ -22,6 +22,8 @@ func (c *Client) Rollback(ctx context.Context, namespace, releaseName string, re
 		return microerror.Mask(err)
 	}
 
+	eventCounter.WithLabelValues(eventName).Inc()
```
tomahawk28, line 25: In this case, if we have a problem during `c.rollback` function call, the counter value would not increase. 
What do you think to call it with `defer` above the rollback line?

tomahawk28 (commented):
Looks good, just one question.

pkg/helmclient/rollback.go
rossf7: Good point. I used an anonymous func so the timer includes counting the rollback. PTAL

tomahawk28: Thanks, maybe not only rollback but install, update also need the same changes, too. :)

rossf7: There are already operatorkit metrics for errors. But OK lets use the same approach for each counter. Done PTAL

tomahawk28 (approved):
LGTM! Thanks for the change.

PR #272, opened 2020-09-21
Fix structs merging error for chart-operator

tomahawk28:
module `github.com/imdario/mergo v0.3.9` has an issue with merging go structs and it results with errors as below from chart-operator side.
```
panic: template: kong-app/templates/deployment.yaml:65:10: executing "kong-app/templates/deployment.yaml" at <include "kong.controller-container" .>: error calling include: template: kong-app/templates/_helpers.tpl:383:4: executing "kong.controller-container" at <include "kong.ingressController.env" .>: error calling include: template: kong-app/templates/_helpers.tpl:238:20: executing "kong.ingressController.env" at <mustMergeOverwrite $autoEnv $userEnv>: error calling mustMergeOverwrite: cannot append two different types (bool, string)
```

Even though we are using the `0.3.8` version on our side, cluster-runtime is using `0.3.9` so it was override. 
```
go mod graph
...
sigs.k8s.io/controller-runtime@v0.5.8 github.com/imdario/mergo@v0.3.9
``` 

Therefore we need another override from our go.mod. 

## Checklist


rossf7:
Wow good find, thanks!

This looks good but don't we need a replace rule? Something like this.

```
// Use latest mergo to fix problem merging Go structs.
replace github.com/imdario/mergo => github.com/imdario/mergo 0.3.11
```

tomahawk28:
@rossf7 Thanks for the suggestion. Although I applied this `replace` directive into helmclient, `chart-operator` always exclude `github.com/imdario/mergo 0.3.11` from go module. This means we need another `replace` directive in any helmclient-applied repository. So I'm inclined to leave it just like this, WDYT?

rossf7:
Yes we'd also need a replace in chart-operator but I think it's better to be explicit about the override.

It's unlikely that go modules will downgrade to 0.3.9 with the bug but if it does there can be a regression. I'm fine with a comment if you think that's better than the replace.

tomahawk28:
Sure, There are not many repositories whom using helmclient directly, but I would like to left this as indirect to make dev process easier.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/giantswarm/microerror v0.2.1
 	github.com/giantswarm/micrologger v0.3.1
 	github.com/google/go-cmp v0.5.2
+	github.com/imdario/mergo v0.3.11 // indirect
```
rossf7, line 13:
```suggestion
        // Use mergo 0.3.11 due to bug in 0.3.9 merging Go structs.
	github.com/imdario/mergo v0.3.11 // indirect
```

What about adding this comment?

tomahawk28: Good idea, I added it, PTAL.

rossf7 (approved):
LGTM!

PR #279, opened 2020-10-01
Adding release name as labels

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13535

## Checklist

### DIFF
--- a/pkg/helmclient/install.go
+++ b/pkg/helmclient/install.go
@@ -16,7 +16,7 @@ func (c *Client) InstallReleaseFromTarball(ctx context.Context, chartPath, names
 
 	t := prometheus.NewTimer(histogram.WithLabelValues(eventName))
 	defer func() {
-		eventCounter.WithLabelValues(eventName).Inc()
+		eventCounter.WithLabelValues(eventName, options.ReleaseName).Inc()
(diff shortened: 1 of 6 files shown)


pkg/helmclient/install.go
```diff
@@ -16,7 +16,7 @@ func (c *Client) InstallReleaseFromTarball(ctx context.Context, chartPath, names
 
 	t := prometheus.NewTimer(histogram.WithLabelValues(eventName))
 	defer func() {
-		eventCounter.WithLabelValues(eventName).Inc()
+		eventCounter.WithLabelValues(eventName, options.ReleaseName).Inc()
```
tomahawk28, line 19: Getting the release name from `options` struct. Not sure why it is in the `options` struct, unlike other commands.

tomahawk28:
Tested in `ginger`

image

pkg/helmclient/install.go
rossf7: That's how its done in the helm action package. Not sure why. Possibly because helm 2 auto generated names but now you have to provide one.

rossf7 (approved):
LGTM

PR #301, opened 2020-12-10
Delete MergeValue function

tomahawk28:
I checked all the usage of `MergeValue` function and almost all repos are updated to `mergo` at the moment. 
We can safely delete this once for all.


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/prometheus/client_golang v1.8.0
 	github.com/spf13/afero v1.4.1
 	github.com/ulikunitz/xz v0.5.7 // indirect
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
tomahawk28, line 20: Not sure why it said indirect, but looks okay.

rossf7: This is just because we used to use it directly. https://github.com/giantswarm/helmclient/blob/17a4ac2f5c5e70a0987dd070dda27150078589aa/pkg/helmclient/merge_values.go#L7

Another dependency must still be using it. Agree should be fine.

rossf7 (approved):
LGTM

PR #315, opened 2021-01-19
Pass SkipCRDs to helmclient

yasn77:
Towards giantswarm/giantswarm#13746

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Pass SkipCRDs to Helm client
+- Pass SkipCRDs to Helm client.
 
 ## [4.1.0] - 2020-12-14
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Pass SkipCRDs to Helm client
```
rossf7, line 10:
```suggestion
- Pass SkipCRDs to Helm client.
```

pkg/helmclient/install.go
```diff
@@ -67,4 +67,5 @@ func (options InstallOptions) configure(action *action.Install, namespace string
 	action.ReleaseName = options.ReleaseName
 	action.Timeout = options.Timeout
 	action.Wait = options.Wait
+	action.SkipCRDs = options.SkipCRDs
```
rossf7, line 70: Nice but the flag also needs to be set in update.go.

yasn77: Not sure why? Since we don't use install flag (`-i`), there isn't a need for skip CRDS.

rossf7: Right, good catch. 👍

rossf7 (approved):
LGTM

PR #320, opened 2021-01-28
 Change GetReleaseHistory to return multiple revisions

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/15213

This is so we can check in chart-operator how many times a release has failed. This prevents too many secrets being created because the max history flag is not respected.

This is a breaking change but GetReleaseHistory is not used. We used to use it with Helm 2 to get version info that was not present in GetReleaseContent.

## Checklist


rossf7:
Tested with chart-operator in giraffe.

tomahawk28 (commented):
At some points, I got errors from helm which indicates it was too big to return as 10 history ledgers are bigger than 2MB.  Maybe we should test this change with prometheus-operator which have a lot of resources?

rossf7:
> At some points, I got errors from helm which indicates it was too big to return as 10 history ledgers are bigger than 2MB. Maybe we should test this change with prometheus-operator which have a lot of resources?

@tomahawk28 Good shout, thanks. I added some debug in GetCurrentState and output the history for all chart CRs on giraffe.

See https://github.com/giantswarm/chart-operator/blob/37baf7474d8cd2cfe45b6e1a2ac8e9fb7ddb19d1/service/controller/chart/resource/release/current.go#L79-L86

It worked fine with all apps including prometheus-operator. So I think this is OK. WDYT?

tomahawk28 (approved):
LGTM, Thanks for pointing out.

PR #341, opened 2021-04-30
Add already exist error

tomahawk28:
## Checklist


rossf7 (commented):
Thanks for looking into this.

I'm just not sure about the name. As it could also be the helm release already exists.

What about `resourceAlreadyExistsError` for the name?

tomahawk28:
You're right, thanks for pointing out! I'm going to change the name as you suggested.

rossf7 (approved):
LGTM

PR #359, opened 2021-08-05
Downgrade Helm to v3.5.4 for compatibility with other libraries

rossf7:
Helm 3.6.X uses k8s 1.21 but we need to use 1.20 for compatibility with CAPI.

## Checklist


rossf7 (commented):
Should have tested with chart-operator before upgrading helm 🤦‍♂️

PR #405, opened 2022-04-11
Align files

architectbot:
This PR is generated by https://github.com/giantswarm/github, it will be merged automatically once all requirements are met


go.mod
```diff
@@ -21,7 +22,7 @@ require (
 )
 
 replace (
-	github.com/containerd/containerd v1.3.4 => github.com/containerd/containerd v1.4.4
+	github.com/containerd/containerd => github.com/containerd/containerd v1.4.4
```
rossf7, line 25: @kubasobon Just FYI I changed the replace here to get rid of the usage of k8s 1.13.

This indirect dependency got pulled in but I think that's OK and means we don't need a custom `.nancy-ignore`.

```
github.com/containerd/continuity v0.3.0 // indirect
```

rossf7, line 25: @kubasobon Just FYI I changed the replace here to get rid of the usage of k8s 1.13.

This indirect dependency got pulled in but I think that's OK and means we don't need a custom `.nancy-ignore`.

```
github.com/containerd/continuity v0.3.0 // indirect
```

rossf7 (approved):

